PR #15, opened 2020-11-02
👷 Improve memory storage

kiyonlin:
- Solve data race in test and make test coverage 100%
- Change `ConfigDefault` to `defaultConfig` and make it private(There is no reason to make it public, right?)
- Make all test cases parallel
- `v.expiry != 0` is cheaper than `v.expiry <= time.Now().Unix()`
- `<-Ticker.C` is already current time(`time.Now()`)

### DIFF
--- a/memory/config.go
+++ b/memory/config.go
@@ -7,16 +7,15 @@ type Config struct {
 	GCInterval time.Duration
 }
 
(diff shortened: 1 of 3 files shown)


kiyonlin:
@Fenny Will we expose all ConfigDefault?

PR #25, opened 2020-11-03
Add DeleteWild to delete by prefix in redis

arham09:
Add DeleteWild function, so it can delete key by pattern.
For example if we have caching key 'category:main', 'category:tree', we can remove the caches by pattern 'category:*'

### DIFF
--- a/redis/redis.go
+++ b/redis/redis.go
@@ -74,6 +74,19 @@ func (s *Storage) Delete(key string) error {
 	return s.db.Del(context.Background(), key).Err()
 }
 
+// DeleteWild key by prefix
+func (s *Storage) DeleteWild(prefix string) error {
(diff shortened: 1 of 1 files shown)


codemicro:
Hi @arham09, thanks for the PR!

The storage drivers in this repository implement the [`fiber.Storage`](https://github.com/gofiber/fiber/blob/ba08653c92f86bc69956b23714f919b705d9381e/app.go#L39-L50) interface, which is used to provide a standard interface for middlewares to store and retrieve data.

Unfortunately, `DeleteWild` is not part of the interface, and as a result, I don't think this PR will be merged. 

Sorry about that!

arham09:
thanks @codemicro , clearly missed the part about the `fiber.Storage` interface in readme, sorry

PR #43, opened 2020-11-24
Add support for MongoDB Atlas

Kalissaac:
This PR adds support for using MongoDB Atlas connections inside Storage.

There is a new boolean option in the config named `Atlas`, and if that is enabled it does the following:
1. Change the protocol to use `mongodb+srv://` (instead of `mongodb://`)
2. Remove the port from the connection string

I didn't see any contribution guidelines, so please let me know if I should change anything here.

### DIFF
--- a/mongodb/mongodb.go
+++ b/mongodb/mongodb.go
@@ -48,7 +48,11 @@ func New(config ...Config) *Storage {
 	if cfg.Username != "" || cfg.Password != "" {
 		dsn += "@"
 	}
-	dsn += fmt.Sprintf("%s:%d", url.QueryEscape(cfg.Host), cfg.Port)
+	if cfg.Atlas == true {
+		dsn += url.QueryEscape(cfg.Host) // Cannot specify port when using MongoDB Atlas
+	} else {
+		dsn += fmt.Sprintf("%s:%d", url.QueryEscape(cfg.Host), cfg.Port)
+	}
 
 	// Set mongo options
 	opt := options.Client()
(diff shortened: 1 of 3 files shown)


hi019:
Thanks for the PR! Can you update the README too?

mongodb/mongodb.go
```diff
@@ -33,7 +33,12 @@ func New(config ...Config) *Storage {
 	cfg := configDefault(config...)
 
 	// Create data source name
-	var dsn = "mongodb://"
+	var dsn = "mongodb"
+	if cfg.Atlas == true {
```
hi019, line 37: We can do `if cfg.Atlas {`

mongodb/mongodb.go
```diff
@@ -43,7 +48,11 @@ func New(config ...Config) *Storage {
 	if cfg.Username != "" || cfg.Password != "" {
 		dsn += "@"
 	}
-	dsn += fmt.Sprintf("%s:%d", url.QueryEscape(cfg.Host), cfg.Port)
+	if cfg.Atlas == true {
```
hi019, line 51: Here too

hi019 (changes requested):

mongodb/mongodb.go
Kalissaac: Ah, didn't know you could do that with Go. Fixed that.

Kalissaac:
> We should also consider an option for the user to provide a query string. It would override the other settings. This would prevent more edge cases like Atlas, but may complicate the code.

Yeah, definitely. I wanted to keep it simple for this PR, but I'd be happy start working on that after.

hi019:
Congrats on merging your first pull request! 🎉 

If you need help or want to chat with us, join us on Discord https://gofiber.io/discord

PR #44, opened 2020-11-26
Added ArangoDB support for storage

antoniodipinto:
This PR adds support for using ArangoDB connection inside Storage.

Respecting the format of all the tests from other storage types, everything is fine.
Please comment here if I need to change something

Thank you

### DIFF
--- a/arangodb/README.md
+++ b/arangodb/README.md
@@ -1,0 +1,107 @@
+# ArangoDB
+A ArangoDB storage driver using `arangodb/go-driver` and [arangodb/go-driver](https://github.com/arangodb/go-driver).
+
+### Table of Contents
+- [Signatures](#signatures)
+- [Installation](#installation)
+- [Examples](#examples)
+- [Config](#config)
+- [Default Config](#default-config)
+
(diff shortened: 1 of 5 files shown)


hi019:
We should consider rejecting PR's that add more storage interfaces to this repository, because we will have to maintain all of these implementations for various databases, some of which we have never used. Instead we could do what [gorilla does](https://github.com/gorilla/sessions#store-implementations) and mention third party stores in the README. 

What do you think @Fenny ?

antoniodipinto:
Hi @hi019 no problem. I understand your point, it's fair and right. 
I will create my own repo of this interface because I really use it in my projects. 

Thank you for your answer

hi019:
Ok. I'll leave this PR open until Fenny replies.

Fenny:
I think we can add popular storage providers as long they are implemented right. By any chance we could add a test environment to the workflow?

Fenny:
I will add a workflow test 👍

antoniodipinto:
Thanks @Fenny and  @hi019!

PR #49, opened 2020-12-01
Add support for MongoDB connection string

Kalissaac:
Adds support for supplying a MongoDB connection URI with `Connection` instead of using the `Host`, `Port`, `User`, and `Password` config options.

Two questions:
1. Is `Connection` the right config name for this?
2. Should the `Atlas` option be removed since Atlas users can now use connection strings instead?

### DIFF
--- a/mongodb/README.md
+++ b/mongodb/README.md
@@ -48,11 +48,25 @@ store := mongodb.New(mongodb.Config{
 	Collection: "fiber_storage",
 	Reset:      false,
 })
+
+// Initialize custom config using connection string
+store := mongodb.New(mongodb.Config{
+	Connection: "mongodb://user:password@127.0.0.1:27017",
+	Database:   "fiber",
+	Collection: "fiber_storage",
+	Reset:      false,
+})
+
 ```
 
 ### Config
 ```go
 type Config struct {
+	// Connection string to use for DB. Will override all other authentication values if used
+	//
+	// Optional. Default is ""
+	Connection string
+
 	// Whether the DB is hosted on MongoDB Atlas
 	//
 	// Optional. Default is false
@@ -98,6 +112,7 @@ type Config struct {
 ### Default Config
 ```go
 var ConfigDefault = Config{
+	Connection: "",
 	Atlas:      false,
 	Host:       "127.0.0.1",
 	Port:       27017,
--- a/mongodb/README.md
+++ b/mongodb/README.md
@@ -41,7 +41,6 @@ store := mongodb.New()
 
 // Initialize custom config
 store := mongodb.New(mongodb.Config{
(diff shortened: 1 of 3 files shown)


hi019:
> Is Connection the right config name for this?
IMO ConnectionURI or ConnectionURL would be better.

> Should the Atlas option be removed since Atlas users can now use connection strings instead?
Yes, since Atlas gives users the exact connection string in their web app.

Fenny:
Good idea, to keep it universal we should also support `ConnectionURL` / `ConnectionURI` for sql drivers.

PS: Not sure which naming convention is correct <https://mongodb.github.io/node-mongodb-native/2.0/tutorials/urls/>, any suggestions?

Kalissaac:
I don't think even Mongo has settled on one specific naming convention since this support article uses URI: https://docs.mongodb.com/manual/reference/connection-string/

My concern is that using `DSN` may be confusing to beginners since the Mongo docs use URL/URI. I do agree that it should be universal though.

Fenny:
I can approve `ConnectionURL`, if you have experience with the SQL storage providers you can also implement it there.

There should be some logic that if `ConnectionURL` is not empty, the other config fields such as `Username` `Database` `Password` etc are not used. If you do not have time to implement this with the sql drivers, you can only implement it for Mongodb and I can add it later to the others 👍

Kalissaac:
> I can approve ConnectionURI, if you have experience with the SQL storage providers you can also implement it there.

Sounds good, I'll add it to Mongo. I haven't used the SQL providers before but I'll take a look at it and see if I can put it in.

> There should be some logic that if ConnectionURI is not empty, the other config fields such as Username Database Password etc are not used.

I believe I added that functionality here? Or were you referring to something else?
https://github.com/Kalissaac/storage/blob/add-connection-string/mongodb/mongodb.go#L38-L53

Kalissaac:
Sorry, there's a typo in the commit name above. The actual code uses `ConnectionURI`.

Fenny:
> > I can approve ConnectionURI, if you have experience with the SQL storage providers you can also implement it there.
> 
> Sounds good, I'll add it to Mongo. I haven't used the SQL providers before but I'll take a look at it and see if I can put it in.
> 
> > There should be some logic that if ConnectionURI is not empty, the other config fields such as Username Database Password etc are not used.
> 
> I believe I added that functionality here? Or were you referring to something else?
> https://github.com/Kalissaac/storage/blob/add-connection-string/mongodb/mongodb.go#L38-L53

Yeah you already did that, nice. Sorry I didn't had time to look at the actual PR yet ( currently on mobile )

Kalissaac:
Gotcha, thanks!

PR #54, opened 2021-03-18
Fix GC deleting values without expiry 

### LINKED ISSUE #53: Postgres - No Expiry not working
When I use postgres storage, I tried to set 0 for no expiry... but the keys are expiring when set to 0

hi019:
Fixes #53


mysql/mysql.go
```diff
@@ -160,16 +160,21 @@ func (s *Storage) Close() error {
 	return s.db.Close()
 }
 
-// Garbage collector to delete expired keys
-func (s *Storage) gc() {
+// gcTicker starts the gc ticker
```
hi019, line 163: I have separated the GC & GC ticker func so that it's easier to test

ReneWerner87 (approved):

Issue #67, opened 2021-03-28
[badger] - Add support to newer badger versions (2.0 or above)


Jictyvoo:
Badger in version 1.6 has some errors under Windows WSL (Linux Subsystem).

So, when trying to running badger on WSL, a panic message is given:
```shell
panic: Mmap value log file. Path=./database/fiber.badger/000000.vlog. Error=exec format error
```

This error, has been fixed in newer versions of badger (2.0 and above), but new versions has no data compability with badger 1.6 versions.

To test this to make sure is a WSL compatibility problem, I run my code in a docker for production, and it has no problems

Jictyvoo:
As I see, Badger 3 is not backward compatible with badger 2, use version 3 them?

PR #68, opened 2021-03-28
badger: update storage to use badgerV3

Jictyvoo:
Implementing a new version of badger to fix problems with WSL, as discussed in #67 

- [x] Add compatibility with Badger-V3
- [x] Add badger config to storage config options
- [x] Check the corresponding of WithTruncate function


Jictyvoo:
I still make some changes, as said in description, to give users more freedom

ReneWerner87:
first of all thanks for the contribution

instead of making a new version in the extra folder, what do you think about replacing the old version ?

Jictyvoo:
> first of all thanks for the contribution
> 
> instead of making a new version in the extra folder, what do you think about replacing the old version ?

The problem with replacing the old version is data in v1 being incompatible with v2, which is also incompatible with v3 But if you see no problem, I can replace instead of creating a new folder

ReneWerner87:
yes no problem, we will mention this in the new release, people should update

Jictyvoo:
Done

badger/badger.go
```diff
@@ -20,7 +20,8 @@ func New(config ...Config) *Storage {
 	cfg := configDefault(config...)
 
 	// Set options
-	opt := badger.DefaultOptions(cfg.Database).WithTruncate(true).WithLogger(nil)
+	opt := cfg.BadgerOptions
+	// TODO see if as a new function similar to WithTruncate(true)
```
ReneWerner87, line 24: please check this TODO

ReneWerner87 (approved):

badger/badger.go
Jictyvoo: Looks like this option or any similar no more exist

The most similar function that I found is [func (Options) WithCompactL0OnClose](https://pkg.go.dev/github.com/dgraph-io/badger/v3@v3.2011.1#Options.WithCompactL0OnClose)

So, I'll remove the TODO

Jictyvoo:
Any more revisions?

PR #74, opened 2021-03-31
🧪 refactor: use `TRUNCATE` to reset the tables for postgres/ mysql

r-52:
MySQL docs: https://dev.mysql.com/doc/refman/8.0/en/truncate-table.html
Postgres docs: https://www.postgresql.org/docs/current/sql-truncate.html

Safes some time if the tables are filled with a lot of data.


ReneWerner87:
https://www.tutorialspoint.com/difference-between-delete-and-truncate-in-sql-query#:~:text=Delete%20and%20truncate%20both%20commands,for%20deleting%20the%20specific%20data.

you are right, good found

no note for all who wonder why it was not made for sqlLite -> here there is only DELETE FROM

PR #76, opened 2021-03-31
:bug: Fix error with no badger options being loaded

Jictyvoo:
My apologies, I forget to test with only Database option given in Config, without using `ConfigDefault`

This PR, is a fix to this bug

When `BadgerOptions` is not passed, a panic error is thrown
```shell
panic: Invalid ValueLogFileSize, must be in range [1MB, 2GB)
```

### DIFF
--- a/badger/.gitignore
+++ b/badger/.gitignore
@@ -1,0 +1,3 @@
+fiber.badger/
+fiber.config.badger/
+fiber.with_options.badger/
--- a/badger/config.go
+++ b/badger/config.go
@@ -68,7 +68,10 @@ func configDefault(config ...Config) Config {
 		cfg.GCInterval = ConfigDefault.GCInterval
 	}
 	overrideLogger := false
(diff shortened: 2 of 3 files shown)


ReneWerner87:
Can we prevent this in the future with an additional test? If so, you are welcome to add one

Jictyvoo:
> Can we prevent this in the future with an additional test? If so, you are welcome to add one

I'll try to add a test later, maybe tomorrow

ReneWerner87:
@Jictyvoo what about the test ?

Jictyvoo:
> @Jictyvoo what about the test ?

Sorry for the delay, I take the weekend to rest a little.
Tests added

ReneWerner87:
Ok, no problem, takr your time

Jictyvoo:
Already added the tests. I don't know if has more to add

ReneWerner87 (approved):

PR #87, opened 2021-04-11
🐛 Fix bug preventing non-utf8 characters being saved in some databases

hi019:
Fixes https://github.com/gofiber/fiber/issues/1258

TODO:
- [x] Write migration instructions
- [x] Create schema check function for Postgres
- [x] Create schema check function for MySQL
- [x] Create schema check function for SQLITE
- [x] Move migration instructions to MIGRATION.md

Postgres: 

```
ALTER TABLE fiber_storage 
ALTER COLUMN v TYPE BYTEA USING v::bytea;
```

MYSQL:
```
ALTER TABLE fiber_storage MODIFY COLUMN v BLOB;
```

### DIFF
--- a/MIGRATE.md
+++ b/MIGRATE.md
@@ -1,7 +1,7 @@
 This document contains instructions for migrating to various storage versions.
 
 ### 0.1 -> 0.2
-v0.2 fixes [a bug](https://github.com/gofiber/fiber/issues/1258) in SQLITE, MYSQL, Postgres and Arangodb in which 
+v0.2 fixes [a bug](https://github.com/gofiber/fiber/issues/1258) in MYSQL, Postgres and Arangodb in which 
 inserting non-UTF8 characters would trigger a panic due to the values being saved in a TEXT column instead of a 
 BYTEA/BLOB column. Migration instructions (note you may need to adjust the table names if you have supplied a custom
 config to the storage):
--- a/MIGRATE.md
+++ b/MIGRATE.md
@@ -1,0 +1,21 @@
+This document contains instructions for migrating to various storage versions.
+
+### 0.1 -> 0.2
+v0.2 fixes [a bug](https://github.com/gofiber/fiber/issues/1258) in SQLITE, MYSQL, Postgres and Arangodb in which 
+inserting non-UTF8 characters would trigger a panic due to the values being saved in a TEXT column instead of a 
+BYTEA/BLOB column. Migration instructions (note you may need to adjust the table names if you have supplied a custom
+config to the storage):
+
+**Postgres**
+```sql
+ALTER TABLE fiber_storage
+ALTER COLUMN v TYPE BYTEA USING v::bytea;
+```
+
+**MYSQL**
+```sql
+ALTER TABLE fiber_storage MODIFY COLUMN v BLOB;
+```
+
+**Arangodb**
+No migration other then updating the library is necessary.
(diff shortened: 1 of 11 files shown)


ReneWerner87 (approved):

MIGRATE.md
```diff
@@ -0,0 +1,21 @@
...
+ALTER COLUMN v TYPE BYTEA USING v::bytea;
+```
+
+**MYSQL**
+```sql
+ALTER TABLE fiber_storage MODIFY COLUMN v BLOB;
+```
+
+**Arangodb**
+No migration other then updating the library is necessary.
```
ReneWerner87, line 21: please add a line for sqlite

mysql/mysql.go
```diff
@@ -23,15 +23,20 @@ type Storage struct {
 }
 
 var (
+	checkSchemaMsg = "The `v` row has an incorrect data type. " +
+		"It should be BYTEA but is instead %s. This will cause encoding-related panics if the DB is not migrated (see https://github.com/gofiber/storage/blob/main/MIGRATE.md). " +
```
ReneWerner87, line 27: change
> It should be BYTEA but is instead %s
to 
> It should be BLOB but is instead %s

mysql/mysql.go
```diff
@@ -178,3 +186,16 @@ func (s *Storage) gcTicker() {
...
+
+func (s *Storage) checkSchema(tableName string) {
+	var data []byte
+
+	row := s.db.QueryRow(fmt.Sprintf(checkSchemaQuery, tableName))
+	if err := row.Scan(&data); err != nil {
+		panic(err)
+	}
+
+	if strings.ToLower(string(data)) != "bytea" {
```
ReneWerner87, line 198: compare with "blob"
> 	if strings.ToLower(string(data)) != "bytea" {
to
> 	if strings.ToLower(string(data)) != "blob" {

sqlite3/sqlite3.go
```diff
@@ -24,15 +23,20 @@ type Storage struct {
 }
 
 var (
+	checkSchemaMsg = "The `v` row has an incorrect data type. " +
+		"It should be BYTEA but is instead %s. This will cause encoding-related panics if the DB is not migrated (see https://github.com/gofiber/storage/blob/main/MIGRATE.md). " +
```
ReneWerner87, line 27: change
> It should be BYTEA but is instead
to
> It should be BLOB but is instead

sqlite3/sqlite3.go
```diff
@@ -170,3 +178,16 @@ func (s *Storage) gcTicker() {
...
+
+func (s *Storage) checkSchema(tableName string) {
+	var data []byte
+
+	row := s.db.QueryRow(fmt.Sprintf(checkSchemaQuery, tableName))
+	if err := row.Scan(&data); err != nil {
+		panic(err)
+	}
+
+	if strings.ToLower(string(data)) != "bytea" {
```
ReneWerner87, line 190: compare with "blob"

mysql/config.go
```diff
@@ -47,6 +47,11 @@ type Config struct {
 	// Optional. Default is 10 * time.Second
 	GCInterval time.Duration
 
+	// CheckSchema checks that the "v" column data type is correct. See https://github.com/gofiber/fiber/issues/1258
+	//
+	// Optional. Default is true
+	CheckSchema bool
```
ReneWerner87, line 53: image
when i add a config, which most do because of their password, the checkSchema switch is transmitted with false

few will consider it, here we need to find another solution

hi019:
LGTM

ReneWerner87:
Thx

PR #153, opened 2021-07-02
Updated Postgre module to use pgx instead of pq

Technerder:
Removed Postgre Close test since the pgx implementation of Close() doesn't return an error,
Removed Storage.Close()'s return method since the pgx implementation of Close() doesn't return an error,
Removed Config.maxIdleConns since pgx implementation doesn't support setting a maximum number of idle connections


hi019:
Thanks for the PR! There are quiet a few breaking changes here. We'll need to release a new major version of the postgres implementation.

Technerder:
Is there anything I could change or do to help?

postgres/README.md
```diff
@@ -16,7 +16,7 @@ func (s *Storage) Get(key string) ([]byte, error)
 func (s *Storage) Set(key string, val []byte, exp time.Duration) error
 func (s *Storage) Delete(key string) error
 func (s *Storage) Reset() error
-func (s *Storage) Close() error
```
ReneWerner87, line 19: please do not change the signature

postgres/postgres.go
```diff
@@ -164,20 +161,21 @@ func (s *Storage) Delete(key string) error {
...
 // Reset all entries, including unexpired
 func (s *Storage) Reset() error {
-	_, err := s.db.Exec(s.sqlReset)
+	_, err := s.db.Exec(context.Background(), s.sqlReset)
 	return err
 }
 
 // Close the database
-func (s *Storage) Close() error {
+func (s *Storage) Close() {
```
ReneWerner87, line 175: please do not change the signature

postgres/postgres_test.go
```diff
@@ -173,7 +174,3 @@ func Test_SslRequiredMode(t *testing.T) {
 		SslMode:  "require",
 	})
 }
-
-func Test_Postgres_Close(t *testing.T) {
```
ReneWerner87, line 177: the store still has a close method, what reason is there for you to remove the test? please adjust it to the new conditions

ReneWerner87 (changes requested):

ReneWerner87:
>  I asked the developer of pgx as to whether or not there was a way to find out if  pgxpool.Pool.Close() fails and they said that it simply blocks until it succeeds like I originally thought. I only removed the Test_Postgres_Close test because I thought it was redundant since the close method can't seem to fail. I also only modified the signature for the storage's Close method for the same reason

please add the test again to test the error free process

> I'd be more than happy to change the signature back and make it return nil for the error but I'm not sure if that'd be considered best practice.

https://github.com/gofiber/fiber/blob/master/app.go#L66
we have several storages, it is important that they all use the same interface, otherwise you cannot use them as intended, please just change the return type

Technerder:
> we have several storages, it is important that they all use the same interface, otherwise you cannot use them as intended, please just change the return type

So I assume adding something like the following would be frowned down upon unless I also added it to all other storage implementations?

```go
func (s *Storage) DB() *pgxpool.Pool {
	return s.db
}
```

ReneWerner87:
No would be okay

Technerder:
Awesome, should I rename the function to ``GetDB`` or just add it in as ``DB``?

hi019:
Thanks for all the work you've done here. If we're going to release a new major version, what would you think of rewriting the config to just accept a pgx connection, as you wrote in your issue? We would be able to remove a lot of code.

Technerder:
I'll start working on changing the module to accept a pgx object instead.

Technerder:
> Thanks for all the work you've done here. If we're going to release a new major version, what would you think of rewriting the config to just accept a pgx connection, as you wrote in your issue? We would be able to remove a lot of code.

Should I remove the postgres config struct altogether? I think it would make sense to do so.

hi019:
Yeah, we should remove the config keys that correspond to the connection configuration.

hi019:
But before you make changes, @ReneWerner87 what do you think of the idea?

ReneWerner87:
passing the connection would be okay for me

Technerder:
I'll start implementing this right now.  I'm thinking of modifying the Config to look somewhat like this
```go
// Config defines the config for storage.
type Config struct {
	// DBPool pgxpool.Pool object
	//
	// Required
	DBPool pgxpool.Pool

	// Table name
	//
	// Optional. Default is "fiber_storage"
	Table string

	// Reset clears any existing keys in existing Table
	//
	// Optional. Default is false
	Reset bool

	// Time before deleting expired keys
	//
	GCInterval time.Duration
}
```

But I'm unsure of how to go about modifying the defaultConfig, should it just panic out if no DBPool object is passed in or should I remove the defaultConfig altogether?

hi019:
@Technerder Yes, panicking should be fine here

Technerder:
Latest commit should allow people to directly pass in a pgxpool.Pool object as opposed to the database host, port, etc

ReneWerner87 (approved):

ReneWerner87:
@Technerder please check 
image
and 
image

Technerder:
@ReneWerner87 I asked the developer of pgx about the dependency ([related issue](https://github.com/jackc/pgx/issues/1046)) and this is what they told me

> My first though was: why does a database driver have a dependency on a JWT library? As it turns out, there isn't any production dependency. It is there because of how modules include all the dependencies including tests in go.sum. It appears it is coming from go-kit. 48f3934 trimmed the go-kit dependency so I think this pseudo-dependency should drop entirely at some point.

Its only being used for some test, not in the actual library code.

ReneWerner87:
@Technerder 
image


and the snyk check failed
image

hi019:
One more thing, can you put the new code in a v2 folder? That way we can issue a new major version with a /v2 import

Technerder:
> @Technerder
> image
> 
> and the snyk check failed
> image

Would you be able to confirm if this is related to CVE-2021-3538?

Technerder:
> One more thing, can you put the new code in a v2 folder (so the path will be `postgres/v2/`)? That way we can issue a new major version with a /v2 import

Should I keep the v1 code there in the base folder?

hi019:
Yep

Technerder:
Verification for "Tech#0067" in joeyxyz's dms

sixcolors:
@ReneWerner87 @Technerder just a comment re the 'security' issue with jackc/pg-con jackc/pg-mock it's a circular dependancy issue that only affects tests. If you check the build deps the vulnerable deps are not included. An issue with how snyk builds and evaluates its dependancy graph (ie build and test dependancies don't show up differently in the go.mod file).

https://github.com/jackc/pgx/issues/1054

rishavs:
Hey folks, checking back on this PR. 
PQ is in maintenance mode now and its maintainers even recommend pgx. https://github.com/lib/pq#status

If this PR is ready, it will be good to merge it as soon as possible, so that we can start using a postgres driver which is in active development.

efectn:
@Technerder pgx v5 seems to be ready for production. Can you test it? The problem should be fixed

Technerder:
@efectn I can test the code usage and functionality by tomorrow but I believe I lack the perms to run a Snyk security check/audit (unless I'm mistaken) and had to ask @ReneWerner87 the last time.

efectn:
Screenshot_20221010-002347_Chrome
Screenshot_20221010-002326_Chrome

Satori uuid seems to be removed. Maybe we need to open an issue for these alerts @Technerder

Technerder:
@efectn would you be able to provide a link to that page? I'm also a little confused about why `yaml.v3` is being flagged since the `go.mod` on the main `pgx` repo shows the current version being used to be `v3.0.1` [as shown here](https://github.com/jackc/pgx/blob/master/go.mod#L19) which should remove the vulnerability according to the screenshot.

efectn:
> @efectn would you be able to provide a link to that page? I'm also a little confused about why `yaml.v3` is being flagged since the `go.mod` on the main `pgx` repo shows the current version being used to be `v3.0.1` [as shown here](https://github.com/jackc/pgx/blob/master/go.mod#L19) which should remove the vulnerability according to the screenshot.

- github.com/gofiber/storage/postgres@0.0.0 › github.com/jackc/pgx@v5.0.2 › golang.org/x/crypto@v0.0.0-20220829220503-c86fa9a7ed90 › golang.org/x/net@v0.0.0-20211112202133-69e39bad7dc2
- github.com/gofiber/storage/postgres@0.0.0 › github.com/jackc/pgx@v5.0.2 › github.com/jackc/puddle@v2.0.0 › github.com/stretchr/objx@v0.4.0 › github.com/stretchr/testify@v1.7.1 › gopkg.in/yaml.v3@v3.0.0-20200313102051-9f266ea9e77c

🤔🤔

efectn:
Created an issue --> https://github.com/jackc/pgx/issues/1352

sixcolors:
From a dependancy CVE standpoint I think we should move forward with this PR review.

Per my comments here https://github.com/jackc/pgx/issues/1352 Snyk go.mod scanning issues are false positives. Checked with govulncheck.

```
~/Documents/GitHub/storage/postgres (pr/153) » govulncheck ./...                                                                                               sixcolors@Jasons-MacBook-Pro
govulncheck is an experimental tool. Share feedback at https://go.dev/s/govulncheck-feedback.

Scanning for dependencies with known vulnerabilities...
No vulnerabilities found.
```

efectn:
All test seems OK but i'm not big fan of adding pgx into the `postgres/v2`. I think it should be another else storage interface named `pgx`. What do you think @Technerder @ReneWerner87

efectn:
> From a dependancy CVE standpoint I think we should move forward with this PR review.
> 
> Per my comments here [jackc/pgx#1352](https://github.com/jackc/pgx/issues/1352) Snyk go.mod scanning issues are false positives. Checked with govulncheck.
> 
> ```
> ~/Documents/GitHub/storage/postgres/v2 (pr/153) » govulncheck .                                                                                            1 ↵ sixcolors@Jasons-MacBook-Pro
> govulncheck is an experimental tool. Share feedback at https://go.dev/s/govulncheck-feedback.
> 
> Scanning for dependencies with known vulnerabilities...
> No vulnerabilities found.
> ```
> 
> And checking which packages are acutally used in v2
> 
> ```
> ~/Documents/GitHub/storage/postgres/v2 (pr/153) » go list -deps -f '{{define "M"}}{{.Path}}@{{.Version}}{{end}}{{with .Module}}{{if not .Main}}{{if .Replace}}{{template "M" .Replace}}{{else}}{{template "M" .}}{{end}}{{end}}{{end}}' | sort -u
> github.com/jackc/pgpassfile@v1.0.0
> github.com/jackc/pgservicefile@v0.0.0-20200714003250-2b9c44734f2b
> github.com/jackc/pgx/v5@v5.0.4
> github.com/jackc/puddle/v2@v2.0.0
> golang.org/x/crypto@v0.1.0
> golang.org/x/text@v0.4.0
> ```

We removed snyk thanks

Technerder:
> All test seems OK but i'm not big fan of adding pgx into the `postgres/v2`. I think it should be another else storage interface named `pgx`. What do you think @Technerder @ReneWerner87

I think removing the older interface with the now deprecated `lib/pq` library would be a good idea since like it's [README](https://github.com/lib/pq/blob/master/README.md#status)suggests it is going to be in maintenance mode. But I think it'd be a breaking change if we just completely removed the current v1 implementation. An alternative way to handle this migration would be to transparently to the user update the v1 package to make use of pgx, that way we could get rid of `lib/pg` entirely and use `pgx` in all implementations,

ReneWerner87:
Somehow we need to declare this as a breaking change and prevent others from automatically updating without being aware of it

2 different storage would only confuse, since from the outside should not necessarily be known about the internal usages and both are postgresql

Unfortunately we push all storages together in one versions number so no differentiation can be made

So my tendency goes to the current v2 implementation

@efectn 
Or we try to make this monorepo multiversionable

efectn:
> Somehow we need to declare this as a breaking change and prevent others from automatically updating without being aware of it
> 
> 2 different storage would only confuse, since from the outside should not necessarily be known about the internal usages and both are postgresql
> 
> Unfortunately we push all storages together in one versions number so no differentiation can be made
> 
> So my tendency goes to the current v2 implementation
> 
> @efectn Or we try to make this monorepo multiversionable

It sounds better to make this repo multiversionable

ReneWerner87:
Then let us doing some bit of research

sixcolors:
Why can’t we just make this API compatible?

I don’t think most users care what package we use behind the scenes.

ReneWerner87:
But it's not possible since the type of storage.Conn() is different

ReneWerner87:
maybe we should use the solutions from https://github.com/awslabs/aws-go-multi-module-repository-tools

ReneWerner87 (approved):

PR #226, opened 2021-10-01
Update CI/CD tests.

efectn:


.github/workflows/test-ristretto.yml
```diff
@@ -8,8 +8,8 @@ jobs:
     strategy:
       matrix:
         go-version:
-          - 1.14.x
-          - 1.15.x
```
hi019, line 12: Why are these two versions being removed?

efectn: https://github.com/gofiber/storage/blob/main/ristretto/go.mod for this. I can try to downgrade version from go.mod

PR #227, opened 2021-10-01
🚀 Add S3 Implementation

pcen:
Adds S3 implementation of the Storage interface


ReneWerner87 (commented):
please add the new storage in the workflow files in the root of the repository under .github

pcen:
Added workflow for s3. Currently there are no unit tests for s3 functionality and it would require a lot of code to mock s3 in unit tests.

hi019:
So if I understand correctly:

You're saying that by avoiding the user having to pass in a specific version of the S3 SDK, this implementation can be updated if the SDK receives improvements (without the user having to change anything)?

tomaswarynyca:
@pcen any news about finishing this?

pcen:
> @pcen any news about finishing this?

The code functionally all works, @hi019 requested changes for some small code cleanup, otherwise all that's left is to somehow test with the GitHub CI, I haven't had time to look into how to do it, but if someone else figures it out before I have time to it's ready to merge

efectn:
I need s3 in one of my projects and this storage can ease my work. So i'll try to finish this PR. 

Btw i'm not sure about SDK version. What about using v2? @pcen

pcen:
> I need s3 in one of my projects and this storage can ease my work. So i'll try to finish this PR.
> 
> Btw i'm not sure about SDK version. What about using v2? @pcen

When I started, I don't think v2 was available. Looking at the Amazon SDK repo for golang, it looks like the latest release is on [v1.42.45](https://github.com/aws/aws-sdk-go/releases/tag/v1.42.45). If you use the latest release version, it shouldn't cause any issues. If there is a newer version of the SDK, until it is implemented in golang, it won't effect us.

efectn:
> > I need s3 in one of my projects and this storage can ease my work. So i'll try to finish this PR.
> > Btw i'm not sure about SDK version. What about using v2? @pcen
> 
> When I started, I don't think v2 was available. Looking at the Amazon SDK repo for golang, it looks like the latest release is on [v1.42.45](https://github.com/aws/aws-sdk-go/releases/tag/v1.42.45). If you use the latest release version, it shouldn't cause any issues. If there is a newer version of the SDK, until it is implemented in golang, it won't effect us.

Amazon has repo for go v2 sdk (https://github.com/aws/aws-sdk-go-v2)

Btw we need to enter env variables(creditianals) for exexuting tests. I think we don't need to add tests because of that. Like https://github.com/gofiber/storage/blob/main/dynamodb/dynamodb_test.go

@ReneWerner87 @hi019 What do you think?

pcen:
> > > I need s3 in one of my projects and this storage can ease my work. So i'll try to finish this PR.
> > > Btw i'm not sure about SDK version. What about using v2? @pcen
> > 
> > 
> > When I started, I don't think v2 was available. Looking at the Amazon SDK repo for golang, it looks like the latest release is on [v1.42.45](https://github.com/aws/aws-sdk-go/releases/tag/v1.42.45). If you use the latest release version, it shouldn't cause any issues. If there is a newer version of the SDK, until it is implemented in golang, it won't effect us.
> 
> Amazon has repo for go v2 sdk (https://github.com/aws/aws-sdk-go-v2)
> 
> Btw we need to enter env variables(creditianals) for exexuting tests. I think we don't need to add tests because of that. Like https://github.com/gofiber/storage/blob/main/dynamodb/dynamodb_test.go
> 
> @ReneWerner87 @hi019 What do you think?

Had no idea about v2! in which case, we should definitely be using it!

ReneWerner87 (changes requested):
please check all workflows, think something needs to be adjusted there, add the new storage to the readme in the root, so this is obvious as storage variant 

and the unittests are still insufficient, they need to be extended, so we can deliver the quality the users are expecting from us

ReneWerner87 (approved):

ReneWerner87 (approved):

s3/s3.go
```diff
@@ -0,0 +1,191 @@
...
+	bucket         string
+}
+
+// New creates a new storage
+func New(config ...Config) *Storage {
+	// Set default config
+	cfg := configDefault(config...)
+
+	// Create s3 session
+	// If config fields of credentials not given, credentials are using from the environment variables (AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY), ~/.aws/credentials, or EC2 instance role.
```
hi019, line 34: We should put this in the readme. Or better yet just link to https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/#specifying-credentials

s3/s3.go
```diff
@@ -0,0 +1,191 @@
...
+	if errors.As(err, &nsk) {
+		return nil, nil
+	}
+
+	return buf.Bytes(), err
+}
+
+// Set key with value
+func (s *Storage) Set(key string, val []byte, exp time.Duration) error {
+	if len(key) <= 0 || len(val) <= 0 {
```
hi019, line 77: What if a user wants to insert an empty file?

hi019 (changes requested):
Thanks for finishing this 👍

PR #272, opened 2021-11-20
fix(github/workflows/security): install gosec fails with package embed is not in GOROOT

### LINKED ISSUE #271: 🐛 bug(github/workflows/security): installing go version < 1.16 with pkg embeds
causes: package embed is not in GOROOT

Need to specify go > 1.16 in go install action.

sixcolors:
security workflow runs gosec which needs go version > 1.16 but install v1.15.15; package embed is not in GOROOT

bumps deps

closes #271


sixcolors:
Not sure what snyk is complaining about, don't have access. Otherwise I'm pretty sure this PR is good.

ReneWerner87 (approved):

sixcolors:
Hey @ReneWerner87 there was also a missing exclude on the root gosec run; fixed with 660a7fs

PR #310, opened 2022-01-25
Public sqlite db options: MaxOpenConns, MaxIdleConns, ConnMaxLifetime.

iredmail:
Default settings of `maxOpenConns` (100) and `connMaxLifetime` is not always good, so please allow to customize them. :)

We set max open connection to `1` to avoid the `SQLITE_BUSY` error.

### DIFF
--- a/sqlite3/README.md
+++ b/sqlite3/README.md
@@ -79,15 +79,18 @@ type Config struct {
 	// //////////////////////////////////
 
 	// MaxIdleConns sets the maximum number of connections in the idle connection pool.
-	// Default is 100.
+	//
+	// Optional. Default is 100.
 	MaxIdleConns int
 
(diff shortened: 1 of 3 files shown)


ReneWerner87:
@iredmail don´t forget the README file
https://github.com/gofiber/storage/blob/main/sqlite3/README.md

and the doc comments above the properties (only a known feature is a good feature)

iredmail:
@ReneWerner87  Updated. :)

ReneWerner87:
@iredmail 

```txt
// //////////////////////////////////
	// Adaptor related config options //
	// //////////////////////////////////

	MaxIdleConns    int
	MaxOpenConns    int
	ConnMaxLifetime time.Duration
	
```
don't really like this block, when it was internal it went through for me, but now that we offer it in config for customization

we should describe each property individually, with default value and effect/use

PR #323, opened 2022-02-04
:fire: Make DynamoDB Production-Ready [🎌 breaking change]

efectn:
- Update aws-sdk-go to v2.
- Add unit tests.
- Make config more detailed.


ReneWerner87 (approved):
we must not forget to mention that it is a breaking change

PR #429, opened 2022-08-08
Add Postgres and MySQL connection strings

Kalissaac:
Adds ability to connect to Postgres and MySQL with connection string as opposed to using the manual config values


ReneWerner87:
Thanks, I will check it in the next few days

ReneWerner87 (approved):

PR #451, opened 2022-08-10
Add Conn() support to all storage drivers.

### LINKED ISSUE #440: [Feature Request] Support for retrieving database connections from storage structures
Provide a way to retrieve the db object from the storage.

Example:
```
store := redis.New()
db := store.DB()
keys := db.keys()
```

Currently theres no way to access the DB from any of the storage drivers, limiting the user options within Fiber.

gaby:
This PR adds support for `Conn()` to all drivers. The main purpose is allow access to the underlying database client from the interface.

Fixes #440


ReneWerner87:
thanks for that big PR -> will check it soon

efectn:
Wouldn't `Conn()` be more logical than `DB()` 🤔🤔

gaby:
@efectn Agree, I can change it

gaby:
I'm still not sure what to do with the `memory` driver, given that it returns a Map of Entry.

gaby:
Done, all functions have been renamed from `DB()` to `Conn()`

arangodb/arangodb_test.go
```diff
@@ -132,3 +132,7 @@ func Test_ARANGODB_Non_UTF8(t *testing.T) {
 func Test_ARANGODB_Close(t *testing.T) {
 	utils.AssertEqual(t, nil, testStore.Close())
 }
+
+func Test_ARANGODB_DB(t *testing.T) {
```
efectn, line 136: Please change Test_*_DB to Test_*Conn_

efectn (changes requested):

arangodb/arangodb_test.go
gaby: Done

bbolt/README.md
```diff
@@ -17,6 +17,7 @@ func (s *Storage) Set(key string, val []byte, exp time.Duration) error
 func (s *Storage) Delete(key string) error
 func (s *Storage) Reset() error
 func (s *Storage) Close() error
+func (s *Storage) Conn() bbolt.DB
```
efectn, line 20: There're some mistakes on READMEs like this. please add pointer if they return pointer

README.md
```diff
@@ -46,6 +46,9 @@ type Storage interface {
 	// Close closes the storage and will stop any running garbage
 	// collectors and open connections.
 	Close() error
+
+	// Returns the underlying DB client
+	Conn() db.Client
```
efectn, line 51: I think we don't need here because of we dont add `Conn()` into the `Storage` interface.

efectn (changes requested):

README.md
gaby: Done

bbolt/README.md
gaby: Done

gaby: The only ones that don't return pointers are `Memory` and `ArangoDB`. I have updated the README for all the other drivers.

efectn (approved):

gaby:
@efectn @ReneWerner87 Who can approve the workflows to be run? Thanks

gaby:
I'm not sure which field to use for ArangoDB.

gaby:
@efectn @ReneWerner87 I replaced the return with `driver.Client` since from the client the user can call `Databases()` and get all the databases, collections, etc.

ReneWerner87 (approved):

PR #476, opened 2022-08-23
Support for automerging dependabot PR's

gaby:
* Adds support for auto-merging minor/patch updates triggered by dependabot. 

Reason for this PR is the high-amount of patch PR's being created by dependabot.

The yaml here is taken from: https://github.com/gofiber/recipes/blob/master/.github/workflows/dependabot_automerge.yml

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,12 +2,12 @@
 
 version: 2
 updates:
-    - package-ecosystem: "github-actions"
-      directory: "/"
-      schedule:
-          interval: daily
-      labels:
-          - "🤖 Dependencies"
+    -   package-ecosystem: "github-actions"
+        directory: "/"
+        schedule:
+            interval: daily
+        labels:
+            - "🤖 Dependencies"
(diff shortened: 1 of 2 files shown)


gaby:
Also can a new release of Storage be minted?

ReneWerner87:
> * Adds support for auto-merging minor/patch updates triggered by dependabot.
> 
> Reason for this PR is the high-amount of patch PR's being created by dependabot.
> 
> The yaml here is taken from: https://github.com/gofiber/recipes/blob/master/.github/workflows/dependabot_automerge.yml

i know we had added it to the reciepe repository, just don't know if it works here

there are no checks in the reciepe repository

don't know if this automerging checks if all unittests etc. are successful

gaby:
@ReneWerner87 I was worried about that too, but their documentation states that it will ONLY auto-merge if the CI/Checks pass.

More information here: https://github.com/dependabot/fetch-metadata#enabling-auto-merge

ReneWerner87:
okay, then lets test it

Issue #147, opened 2021-06-26
[Feature Request] Ability to pass in or retrieve database connections to/from storage structures


Technerder:
Like the title suggests, I believe the should be a able to either pass in a database connection to session storage structures or be able to be able to retrieve them for other non session specific database operations as opposed to having 2 separate/different sql drivers and connections (1 for sessions and 1 for other db stuff)

anton7r:
@Technerder so you propose that you could pass a db interface instead of the config to the storage middleware and then later on also be able to retrieve it?

Sounds like a good idea, might start working on a PR pretty soon.

Technerder:
Yeah, working on something like this here https://github.com/gofiber/storage/pull/153

anton7r:
https://github.com/gofiber/storage/pull/153/files#diff-a537808f48341a30aa18ae06a27a21209fa58e4f2a0247dfb933edecce97a12eR13

i see, perhaps i should follow the same sort of design philosophy with the other storage interfaces with a v2 subfolder on the other ones.

gaby:
@Technerder I think this is fixed by #451

gaby:
This is now supported by calling: 

```go
store.Conn() // returns pointer to DB connection used by the store
```

PR #547, opened 2022-10-10
[Feat] Allow reuse of an existing db connection when creating a MySQL storage driver

DGivney:
Support for passing an existing db connection was requested in this thread: https://github.com/gofiber/storage/issues/147

This pull request implements support for this feature in the MySQL storage driver, allowing the developer to optionally reuse their existing db connection to Get/Set session data.

It also updates the README removing references to postgres.

### DIFF
--- a/mysql/README.md
+++ b/mysql/README.md
@@ -51,8 +51,16 @@ store := mysql.New(mysql.Config{
 })
 
 // Initialize custom config using connection string
-store := postgres.New(postgres.Config{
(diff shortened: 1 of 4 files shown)


ReneWerner87:
- [ ] please add a test case for your adjustment

DGivney:
Hi Rene

There were no test cases for the existing functionality of the 'New' method so I wasn't sure what would be the best way to test this.  

This test case creates the 3 different ways of returning a db connection in the interface (through authentication values or ConnectionURI or by passing an existing connection).

Issue #591, opened 2022-10-28
🐛 memcache utils.Trim undefined latest release


sixcolors:
Using go.mod with:
```
github.com/gofiber/fiber/v2 v2.38.1
github.com/gofiber/storage/memcache v0.0.0-20221005114327-4e6f1c95ff26
```
When trying to run I get
```
# github.com/gofiber/storage/memcache
../go/pkg/mod/github.com/gofiber/storage/memcache@v0.0.0-20221027071415-dca8f183e44b/memcache.go:24:36: undefined: utils.Trim
```

if I change memcache.go line 23 from  `serverList := strings.Split(utils.Trim(cfg.Servers, ' '), ",")` to  `serverList := strings.Split(strings.Trim(cfg.Servers, " "), ",")` it works.

@efectn mentioned utils.Trim was used for performance on older go versions. It seems that this is broken with the latest goofier release when using go 1.19 in my package however.

sixcolors:
utils.Trim is used in memcache.go in one place only `func New(config ...Config) *Storage`

Performance of strings.Trim should not be a concern in this case.

ReneWerner87:
In my experience, it would be better to find out the cause of the problem than to work around it, because other places also use these functions.

Will try it myself tomorrow, but the error looks like it has something to do with the modules and not the function itself.

sixcolors:
> In my experience, it would be better to find out the cause of the problem than to work around it, because other places also use these functions.
> 
> Will try it myself tomorrow, but the error looks like it has something to do with the modules and not the function itself.

It does appear to be an issue with modules. The utils.Trim function has been removed from the current utils package, and it seems that the older version is not working.

go.mod for memcahce package is at go 1.14 while fiber is 1.19

ReneWerner87:
But Version https://github.com/gofiber/storage/blob/main/memcache/go.mod#L7
0.1.2

exists https://github.com/gofiber/utils/releases/tag/v0.1.2

with the Trim function https://github.com/gofiber/utils/blob/v0.1.2/strings.go#L39

sixcolors:
> But Version https://github.com/gofiber/storage/blob/main/memcache/go.mod#L7 0.1.2
> 
> exists https://github.com/gofiber/utils/releases/tag/v0.1.2
> 
> with the Trim function https://github.com/gofiber/utils/blob/v0.1.2/strings.go#L39

I see that; something is deffinatly not working as expected.

ReneWerner87:
[...]

PR #592, opened 2022-10-28
fix: utils.Trim undefined latest release go 1.19

sixcolors:
fixes: #591


ReneWerner87 (changes requested):
Will try it myself tomorrow, but the error looks like it has something to do with the modules and not the function itself.

We should find the cause of this and not work past it, because other places and packages work with the same functionalities

PR #604, opened 2022-11-01
Add support to GORM

gandaldf:
I hope this can be useful to the project.
Let me know if I need to fix anything!


ReneWerner87:
thank you for your work

we will consult internally whether to adopt the pull request

gandaldf:
Thank you! The motivation behind this PR is to handle sessions just as you handle the rest of your data.
Without this feature you have to check the GORM dialect in use and manually select the single store package to handle your sessions; when you have to support 4 RDBMS it could be pretty annoying.

gorm/gorm.go
```diff
@@ -0,0 +1,222 @@
...
+		case <-s.done:
+			return
+		case t := <-ticker.C:
+			s.gc(t)
+		}
+	}
+}
+
+// gc deletes all expired entries
+func (s *Storage) gc(t time.Time) {
```
efectn, line 220: I think we can use same logic here. https://github.com/gofiber/storage/blob/main/memory/memory.go

New logic is more performant

gorm/gorm.go
```diff
@@ -0,0 +1,222 @@
...
+		// Use passed db
+		db = cfg.Db
+
+		sqldb, err = db.DB()
+		if err != nil {
+			panic(err)
+		}
+	} else {
+		// Create db
+		switch cfg.Dialect {
```
li-jin-gou, line 50: I'm wondering if we just need to support DB parameters, as adding other parameters isn't necessarily useful.

gandaldf: Well, I think that every Storage package should only support a DB parameter initialized outside the package itself.
Many DBs have a lot config options that make it hard to fully support them, other times you just want to share your already initialized data connection for sessions too.
I'm not experienced enough to discuss precedent decisions, so I have just tried to copy the same code already in production but if you need it I can change it...
Maybe I can find some time to align all other Storage packages to support just an external connection pointer.

gorm/gorm.go
gandaldf: Ok! I can change it... but we should change it in MySQL and all other Storage packages.

ReneWerner87:
the storages should only be hardware close driver and not contain an intermediate layer like an orm model

we made a small voting internally about this and came to the result that we don't merge this pull request for the time being until the community decides otherwise

PR #617, opened 2022-11-14
Fix and optimize memory storage

leonklingele:


leonklingele:
Taken from https://github.com/gofiber/fiber/pull/2207

ReneWerner87 (approved):

PR #636, opened 2022-11-20
Deprecate support for go 1.14 and go 1.15

gaby:
* Update all go.mod files to use `go 1.19`
* Removed `go 1.14 and go 1.15` from all the test cases
* Added `go 1.18` to all the drivers tests.

Note:
* I had to remove `go 1.16` from the test matrix in `test.yml` (local storage). The reason is because badger no longer supports 1.16 as seen in #608 

Fixes #608 #548


efectn:
I think we can split .github/workflows/test.yml tests

gaby:
> I think we can split .github/workflows/test.yml tests

Sounds good! I'd add that to this PR

.github/workflows/test-sqlite3.yml
```diff
@@ -0,0 +1,49 @@
+on:
+    push:
+        branches:
+            - master
+            - main
+    pull_request:
+name: "Tests Local Storage"
```
ReneWerner87, line 7: don't forget to update the badges links in the readme in the root folder

efectn (approved):

.github/workflows/test-badger.yml
```diff
@@ -0,0 +1,44 @@
+on:
+    push:
+        branches:
+            - master
+            - main
+    pull_request:
+name: "Tests Badger"
```
ReneWerner87, line 7: don't forget to update the badges links in the readme in the root folder

PR #659, opened 2022-12-11
Draft: Use channels to sync ristretto tests

gaby:
I noticed recently that the tests for ristretto fail randomly because the expected values is not there, even if we haven't made any changes to the code/tests. Further checking, it seems the issues is caused by using time.Sleep(10000) during the tests which is not reliable.

This PR updates the tests to use channels and synchronization to wait for the expected value to be set in the store before deleting the key and/or checking the result. This eliminates the need for the time.Sleep call, which can make the test slower and less reliable. The test will wait for the expected result or a timeout to occur, and will fail if the timeout occurs before the expected result is available. This can make the test faster and more reliable.


gaby:
It seems even the official ristretto tests are using time.Sleep for 10 ms.

Source: https://github.com/dgraph-io/ristretto/blob/main/cache_test.go#L357

There's an open PR that may be related to the issue we see in the tests: https://github.com/dgraph-io/ristretto/pull/286

gaby:
In the comments they suggested migrating to https://github.com/outcaste-io/ristretto which fixes the issues from https://github.com/dgraph-io/ristretto/pull/286

ReneWerner87 (approved):

ReneWerner87:
image
https://github.com/gofiber/storage/actions/runs/4124801449/jobs/7124673103

ReneWerner87:
not 100% stable i think
i run it 3 times and it happens

gaby:
@ReneWerner87 Yeah, even with these changes it's not stable. Thoughts on using the fork https://github.com/outcaste-io/ristretto ?

gaby:
> whats the difference to the main repo?

It has the changes from this PR: https://github.com/dgraph-io/ristretto/pull/286

PR #174, opened 2021-07-27
[Feat] allow for usage of redis FailoverClient with Sentinel

binhnguyenduc:
We often use Redis Sentinel on Production environment for a more robust and highly available set up. This PR allows for configuration of Redis Sentinel on redis Storage Adapter and will not break any existing code.

Please consider merging this PR as some people might find this useful as well. 

Thank you! 

Changes:

allow config + init of redis FailoverClient with Sentinel to provide a more robust installation of Redis on Production environment.

- config.go: add relevant configuration options for Sentinel. These options will have not break existing code that depends on this library, if not configured explicitly.
- redis.go: add a step to init redis db with redis.NewFailoverClient() if EnableFailover=true (default false). To avoid breaking existing code, URL still takes precedent and only used to config Redis Standalone (i.e. cannot config Sentinel with URL as that would change the string format)

PS: Please let me know if anything needs revising.

PR #772, opened 2023-04-09
Support for Redis Universal Client

### LINKED ISSUE #382: [Question] support for redis cluster client
the integration of the middleware for redis is assuming that i'm using a regular client, is it even possible to define a ClusterClient with a set of addresses? am i missing something?

EDIT:
nevermind, i made an interface adapter for the *redisClusterClient type,
maybe i will make a PR after some proper tests.

gaby:
Added support for Redis Universal Client which is a wrapper that allows using a Redis Client, Redis Failover Client, and Redis Cluster Client through the same interface.

More information here: https://redis.uptrace.dev/guide/universal.html

**Note**: _This is an interface breaking change since now we return `redis.UniversalClient` instead of `*redis.Client`, this will require releasing this module under v2.x.x_
Follow-on from #174
Fixes #382


gaby:
As expected running Sentinel requires multiple Redis instances, which are not supported by the `actions-setup-redis` action.

gaby:
I'm going to update the logic to use the universal client, that one works for Cluster, Single, Sentinel.

https://pkg.go.dev/github.com/redis/go-redis/v9#NewUniversalClient

gaby:
Using the `UniversalClient` is a massive breaking change, since it requires getting rid of `redis.Client` and changing the store to `db *redis.UniversalClient`. But it's the only clean way of supporting `Client`, `SentinelClient` and `ClusterClient` all at once.

Just let me know, and I will do the changes else the only solution for now is not to test the configuration fields of the Sentinel client.

gaby:
The other option would be to migrate to the redis.UniversalClient, which will affect the Conn() return type. Then releasing the next version of this module under v2.x.x

efectn:
> The other option would be to migrate to the redis.UniversalClient, which will affect the Conn() return type. Then releasing the next version of this module under v2.x.x

Seems better to me

efectn (commented):
Don't forget to make it `v2`

gaby:
> Don't forget to make it `v2`

Done!

gaby:
> @gaby tests only connect to a single Redis instance, should we actually use a cluster in the Tests?

Correct. I based the New() using the documentation from Redis. Unfortunately I havent found a github action for running a Redis Cluster.

I added the extra tests to make sure the behavior is working as expected.

sixcolors (approved):

gaby:
@sixcolors https://github.com/pfapi/redis-cluster-service found this one from 2022 that I can give a try.

sixcolors:
> @sixcolors https://github.com/pfapi/redis-cluster-service found this one from 2022 that I can give a try.

Looks good. Do we keep current action for TLS and single server connections and add this for cluster?

gaby:
> > @sixcolors https://github.com/pfapi/redis-cluster-service found this one from 2022 that I can give a try.
> 
> Looks good. Do we keep current action for TLS and single server connections and add this for cluster?

Yes, it would just run in the background until the cluster test runs at the end. By default it uses very high ports 7000-7005 so there shouldnt be conflicts. I'm making the changes in like 15mins

sixcolors:
> @sixcolors We got Redis Cluster tests now :-)

Awesome work @gaby. Thanks!

sixcolors (approved):

ReneWerner87 (approved):

PR #783, opened 2023-04-16
🚀 [Feature]: Split fiber.Storage to multiple interfaces

carmeloriolo:
Fixes https://github.com/gofiber/fiber/issues/2399

See https://github.com/gofiber/fiber/pull/2418

### DIFF
--- a/storage.go
+++ b/storage.go
@@ -5,22 +5,43 @@ import "time"
 // Storage interface for communicating with different database/key-value
 // providers. Visit https://github.com/gofiber/storage for more info.
 type Storage interface {
+	StorageGetter
+	StorageSetter
+	StorageDeleter
+	StorageResetter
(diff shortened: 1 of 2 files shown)


storage.go
```diff
@@ -5,22 +5,43 @@ import "time"
 // Storage interface for communicating with different database/key-value
 // providers. Visit https://github.com/gofiber/storage for more info.
 type Storage interface {
+	StorageGetter
```
gaby, line 8: Do we need to update the interface on each Storage driver?

gaby (changes requested):

storage.go
gaby: Nvmd, it's a struct :-)

gaby (approved):
LGTM

PR #851, opened 2023-06-15
security checks, file access rights and error handling fixed

berkantay:
- Security vulnerabilities fixed.
- File access right changed.
- Error handling added on file validation check.

### DIFF
--- a/pebble/pebble.go
+++ b/pebble/pebble.go
@@ -132,6 +132,11 @@ func isValid(fp string) bool {
 		return false
 	}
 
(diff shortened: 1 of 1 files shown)


berkantay:
gosec version may not be the updated one i believe that is the reason why it fails on security stage

gaby (approved):

gaby:
@ReneWerner87 these changes are needed for #853

ReneWerner87 (approved):

PR #920, opened 2023-08-09
Add Oracle Coherence storage driver

tmiddlet2666:
This pull request aims to add Coherence implementation into storage.
Thanks for the review, Please let me know of any issues to resolve.


gaby:
@tmiddlet2666 Thanks for the PR, lets us know when its ready for review

tmiddlet2666:
> @tmiddlet2666 Thanks for the PR, lets us know when its ready for review

Thanks, not sure why the test is vailing on downloading our coherence-ce image https://github.com/gofiber/storage/actions/runs/5804095386/job/15734559701?pr=920

Getting `Error: Process completed with exit code 56.`

The image is around 257MB, so reasonably large. Are there any limits on disk or downloads?

gaby:
> > @tmiddlet2666 Thanks for the PR, lets us know when its ready for review
> 
> Thanks, not sure why the test is vailing on downloading our coherence-ce image https://github.com/gofiber/storage/actions/runs/5804095386/job/15734559701?pr=920
> 
> Getting `Error: Process completed with exit code 56.`
> 
> The image is around 257MB, so reasonably large. Are there any limits on disk or downloads?

250MB is not large by any means. The problem is something on that health/start script. Have you tried running that shell code locally?

coherence/coherence.go
```diff
@@ -0,0 +1,128 @@
...
+	ctx        context.Context
+}
+
+// Config defines configuration options for Coherence connection.
+type Config struct {
+	// Address to connect to, defaults to "localhost:1408"
+	Address string
+
+	// UseSSL specified if to use SSL or plain text, defaults to false
+	UseSSL bool
```
gaby, line 30: This should be a tls Config, like the other drivers. Else it won't work for private/self-signed/corporate environments

coherence/coherence.go
```diff
@@ -0,0 +1,128 @@
...
+type Storage struct {
+	session    *coh.Session
+	namedCache coh.NamedCache[string, []byte]
+	ctx        context.Context
+}
+
+// Config defines configuration options for Coherence connection.
+type Config struct {
+	// Address to connect to, defaults to "localhost:1408"
+	Address string
```
gaby, line 27: Split the host and port into separate fields, or rename this to "hostname"

coherence/coherence.go
```diff
@@ -0,0 +1,128 @@
...
+
+	// UseSSL specified if to use SSL or plain text, defaults to false
+	UseSSL bool
+
+	// SessionTimeout is the default session timeout to connect to Coherence, defaults to 30s
+	SessionTimeout time.Duration
```
gaby, line 33: Rename to "timeout"

coherence/coherence.go
```diff
@@ -0,0 +1,128 @@
...
+
+	// SessionScope defines a scope allowing for multiple storage sessions
+	SessionScope string
```
gaby, line 36: Rename to scope or scopename

coherence/coherence.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return newCoherenceStorage(session, scopeName)
+}
+
+// newCoherenceStorage returns a new Coherence [Storage].
+func newCoherenceStorage(session *coh.Session, cacheName string) (*Storage, error) {
+	nc, err := coh.GetNamedCache[string, []byte](session, "fiber$"+cacheName)
```
gaby, line 89: Can we make this "fiber$" prefix configurable? In case someone already has a DB with data, that prefix wouldnt let them use it, right?

coherence/README.md
```diff
@@ -0,0 +1,94 @@
...
+- [Default Config](#default-config)
+
+### Signatures
+```go
+func New(config ...Config) (*Storage, error)
+func (s *Storage) Get(key string) ([]byte, error)
+func (s *Storage) Set(key string, val []byte, exp time.Duration) error
+func (s *Storage) Delete(key string) error
+func (s *Storage) Reset() error
+func (s *Storage) Close() error
```
gaby, line 19: The `Conn()` is missing

coherence/coherence_test.go
```diff
@@ -0,0 +1,181 @@
...
+	val, err = testStore2.Get(key1)
+	utils.AssertEqual(t, err, nil)
+	utils.AssertEqual(t, value2, val)
+
+	utils.AssertEqual(t, testStore1.Close(), nil)
+	utils.AssertEqual(t, testStore2.Close(), nil)
+}
+
+// newTestStore returns a new Coherence Store and ensures it is reset.
+func newTestStore(t *testing.T, config ...Config) (*Storage, error) {
```
gaby, line 174: Instead of this, we should add a "Reset" boolean to the config. To allow resetting during creation. It should default to False.

.github/workflows/test-coherence.yml
```diff
@@ -0,0 +1,48 @@
...
+        go-version:
+          - 1.19.x
+          - 1.20.x
+    steps:
+      - name: Fetch Repository
+        uses: actions/checkout@v3
+      - name: Startup Coherence
+        run: |
+          docker run -d -p 1408:1408 -p 30000:30000 ghcr.io/oracle/coherence-ce:22.06.5
+          counter=0
```
gaby, line 26: My guess is that his doesnt run. It has to be done in a separate step.

```
- name: Wait for Database
  run: |
    #YourHealthCheckHere
```

coherence/coherence.go
gaby: Nvmd this comment :-)

coherence/coherence.go
tmiddlet2666: The `Config.ScopeName` is exactly for that reason. (The fiber$ prefix is just for the cache name so that it wouldn't )

If you are for a default coherence Store, the underlying coherence cache will be "fiber$default-store" (which i've changed from fiber-session-store as its not just for sessions).

If this is asked for by difference processes or new vars, they will map to the same underlying coherence cache.

If you want separate stores, you can use
`store1, err := New(Config{ScopeName: "scope1"})`  - which will store data in coherence cache named "fiber$scope1"

`store2, err := New(Config{ScopeName: "scope2"})` - which will store data in coherence cache named "fiber$scope2"

I guess the scopeName is similar to the `Bucket` in couchbase, but we also add a prefix to avoid conflict with other coherence caches.

Hope that makes sense.

.github/release-drafter-coherence.yml
```diff
@@ -0,0 +1,50 @@
...
+            - 'patch'
+            - '📒 Documentation'
+            - '☢️ Bug'
+            - '🤖 Dependencies'
+            - '🧹 Updates'
+    default: patch
+template: |
+    $CHANGES
+
+    **Full Changelog**: https://github.com/$OWNER/$REPOSITORY/compare/$PREVIOUS_TAG...memcache/v$RESOLVED_VERSION
```
ReneWerner87, line 48: change the package name in this line

tmiddlet2666: thanks, done and will push

coherence/coherence.go
tmiddlet2666: Currently we only support setting these via environment variables as described here - https://pkg.go.dev/github.com/oracle/coherence-go-client/coherence#NewSession
I'm in the process of changing this, but may be able to do a workaround in the meantime.

gaby: That's fine, else it can be added as a feature after initial release.

gaby:
Adding TLS issue for refenrece here: https://github.com/oracle/coherence-go-client/issues/47

@tmiddlet2666 Overall it looks good so far, do you want to add go1.21 to the CI now that it was released yesterday?

tmiddlet2666:
Thanks @gaby 
I have added the Config options and currently they set env variables, but when the above issue in coherence-go-client is fixed and released, i can update this package.
Will add go 1.21.

gaby:
> Thanks @gaby I have added the Config options and currently they set env variables, but when the above issue in coherence-go-client is fixed and released, i can update this package. Will add go 1.21.

For reference for the Redis driver, we allow the user to provide a `tls.Config` object. That way they can customize it to their needs. 

https://github.com/gofiber/storage/blob/main/redis/redis.go#L50C4-L50C4

tmiddlet2666:
@gaby - have moved from pending to proper PR

gaby:
> @gaby - have moved from pending to proper PR

Thanks for the contribution, will start reviewing this week. So far it looks good.

gaby (approved):
lgtm, thank you!

ReneWerner87 (approved):

ReneWerner87:
@tmiddlet2666 @gaby 
image

tmiddlet2666:
> @tmiddlet2666 @gaby image what is the best thing to do in this case?

Let me work on adding the *tls.Config support. 
I'll need to release a new go client.
will take me a couple of days and i'll ping you.
Thanks

gaby:
> > @tmiddlet2666 @gaby image what is the best thing to do in this case?
> 
> Let me work on adding the *tls.Config support. I'll need to release a new go client. will take me a couple of days and i'll ping you. Thanks

Sounds like a plan, thank you!

PR #931, opened 2023-08-15
Add support for rueidis storage driver

gaby:
- Adds support for the fast redis storage driver known as `rueidis`.
- This alternative official library supports Client Side Caching, Auto Pipelining, Generics OM, RedisJSON, RedisBloom, RediSearch, etc.

Client: https://github.com/redis/rueidis


rueidis/rueidis.go
```diff
@@ -0,0 +1,113 @@
+package rueidis
+
+import (
+	"context"
+	"time"
+
+	"github.com/gofiber/fiber/v2/utils"
```
efectn, line 7: why not to use github.com/gofiber/utils here

rueidis/rueidis_test.go
```diff
@@ -0,0 +1,220 @@
...
+
+	err := store.Set(key, val, 0)
+	utils.AssertEqual(t, nil, err)
+
+	result, err := store.Get(key)
+	utils.AssertEqual(t, nil, err)
+	utils.AssertEqual(t, val, result)
+
+	err = store.Delete(key)
+	utils.AssertEqual(t, nil, err)
```
efectn, line 217: I think we can use testify instead of utils.AssertEqual

gaby: Agree, actually all the drivers should use `testify`

rueidis/rueidis.go
gaby: True, will change it in a bit.

gaby:
I'm going to remove those TLS certs today. The Redis GH Action has support for generating ones

rueidis/rueidis.go
gaby: Done, using `utils/v2` now

rueidis/rueidis_test.go
gaby: Done

rueidis/rueidis_test.go
```diff
@@ -0,0 +1,217 @@
+package rueidis
+
+import (
+	"crypto/tls"
+	"log"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
```
efectn, line 9: Should we use require or assert

gaby: It will make the test end faster during failure, so sure. I can fix it

gaby:
> I'm going to remove those TLS certs today. The Redis GH Action has support for generating ones

^This is more complicated that I thought, the directory for the certs is random per Job. So i'm going to leave it the way it is

rueidis/rueidis_test.go
gaby: Done

efectn (approved):

rueidis/tests/gen-test-certs.sh
```diff
@@ -0,0 +1,57 @@
...
+
+# Generate some test certificates which are used by the regression test suite:
+#
+#   ./tls/ca.{crt,key}          Self signed CA certificate.
+#   ./tls/redis.{crt,key}       A certificate with no key usage/policy restrictions.
+#   ./tls/client.{crt,key}      A certificate restricted for SSL client usage.
+#   ./tls/server.{crt,key}      A certificate restricted for SSL server usage.
+#   ./tls/redis.dh              DH Params file.
+
+generate_cert() {
```
ReneWerner87, line 11: how long are these certificates valid and do we really need to check them into git? 
if so i would like to move them to the .github folder
maybe you can use them for other adapters too

.github/workflows/gosec.yml
```diff
@@ -35,11 +35,7 @@ jobs:
           cache: false
       - name: Install Gosec
         run: go install github.com/securego/gosec/v2/cmd/gosec@latest
-      - name: Run Gosec (root)
-        working-directory: .
-        run: |
-          gosec -exclude-dir=arangodb -exclude-dir=badger -exclude-dir=couchbase -exclude-dir=dynamodb -exclude-dir=etcd -exclude-dir=memcache -exclude-dir=memory -exclude-dir=mongodb -exclude-dir=mysql -exclude-dir=postgres -exclude-dir=redis -exclude-dir=ristretto -exclude-dir=sqlite3 -exclude-dir=s3 -exclude-dir=bbolt -exclude-dir=azureblob -exclude-dir=mssql -exclude-dir=pebble ./....
```
ReneWerner87, line 41: instead of removing this we could just run it for root or ? with "." at the end
`gosec .`
then we don't have to customize anything in the future and it still checks the code in the root

ReneWerner87 (changes requested):
some comments to answer

rueidis/tests/gen-test-certs.sh
gaby: These are the same certs from the Redis Storage. We could move them somewhere elsd, for reusing

.github/workflows/gosec.yml
gaby: We can run `gosec .` since it gets confused with all the go.mod files.

ReneWerner87: but locally it worked for me
it then takes only the root folder and no subfolders

rueidis/tests/gen-test-certs.sh
ReneWerner87: ok can you move this then to the .github

how long are these valid?

.github/workflows/gosec.yml
gaby: Done

rueidis/tests/gen-test-certs.sh
gaby: Certs are now dynamically generated during CI

ReneWerner87 (approved):

PR #979, opened 2023-09-07
Fix incorrect default timeout value for Coherence storage driver

tmiddlet2666:
This PR is to change the default timeout which was incorrectly set to 30 millis instead of 30 seconds.


ReneWerner87 (approved):

tmiddlet2666:
thanks @ReneWerner87 !

PR #982, opened 2023-09-10
Support for using options.TLSConfig if returned by redis.ParseURL()

### LINKED ISSUE #981: 🐛 [Bug]: Redis secure connection not working properly
### Bug Description

When trying to connect to a rediss (redis secure) it fails.

I'm supposing that is a problem with the storage/redis, since go-redis is handling it [here](https://github.com/redis/go-redis/blob/c3098d5f7eba88330638a8b2b3e23d42cd5c9338/options.go#L302-L307) and returning options with tls, but storage/redis is not handling it as we can see it [here](https://github.com/gofiber/storage/blob/redis/v2.0.3/redis/redis.go#L26-L35), it's not setting the TLS config.

### How to Reproduce

1. Create a new redis storage
2. Pass the URL as rediss (redis secure) protocol
3. Try to use the store

### Expected Behavior

Connection established with TLS.

### Storage package Version

v2.0.3

### Code Snippet (optional)

```go
package main

import "fmt"

func main() {

	store = redis.New(redis.Config{
		URL: "rediss://<user>:<pass>@127.0.0.1:6379/<db>",
	})

	fmt.Println(store.Get(key))

}
```

### Checklist:

- [X] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [X] I have checked for existing issues that describe my problem prior to opening this one.
- [X] I understand that improperly formatted bug reports may be closed without explanation.

gaby:
When using `rediss://<user>:<pass>@127.0.0.1:6379/<db>` the Redis.Options will return a `options.TLSConfig` which we are ignoring causing Redis Secure connections to not work properly.

- Removed old scripts to generate TLS certs.
- Added subAltName to be able to verify against `localhost`.
- Added new test with TLS verification enabled.
- Added extra instance of Redis that disables MTLS to be able to test that  `rediss://<user>:<pass>@127.0.0.1:6379/<db>` works correctly.

Fixes #981


gaby:
I can't get Go to accept the cert even though they are valid according to OpenSSL

redis/redis_test.go
```diff
@@ -146,7 +146,7 @@ func Test_Redis_Initalize_WithURL(t *testing.T) {
 }
 
 func Test_Redis_Initalize_WithURL_TLS(t *testing.T) {
-	cer, err := tls.LoadX509KeyPair("./tests/tls/client.crt", "./tests/tls/client.key")
+	cer, err := tls.LoadX509KeyPair("/home/runner/work/storage/storage/tls/client.crt", "/home/runner/work/storage/storage/tls/client.key")
```
efectn, line 149: Why don't you use relative path here

gaby: It's easier to read rather than "../../../path/to/cerr"

efectn: > It's easier to read rather than "../../../path/to/cerr"

But it won't work on local machines

gaby: I have never been able to run these tests locally. I can change it if needed, the certs are no longer just for Redis, they can be used by other drivers

ReneWerner87 (approved):

Issue #968, opened 2023-08-31
🚀 [Feature]: Add benchmarks for storage drivers


efectn:
### Feature Description

We have to add benchmarks to check performance changes in PRs and commits. Otherwise we won't have an information about performance issues.

### Code Snippet (optional)

```go
package main

import "github.com/gofiber/storage/%package%"

func main() {
  // Steps to reproduce
}
```

### Checklist:

- [X] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [X] I have checked for existing issues that describe my suggestion prior to opening this one.
- [X] I understand that improperly formatted feature requests may be closed without explanation.

PR #992, opened 2023-09-14
Add benchmarks for storage drivers

efectn:
closes https://github.com/gofiber/storage/issues/968

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -19,6 +19,64 @@ name: Benchmark
 jobs:
   Compare:
     runs-on: ubuntu-latest
+    services:
+      arangodb:
+        image: 'arangodb:latest'
+        env:
+          ARANGO_NO_AUTH: 1
+        ports:
+          - '8529:8529'
+      dynamodb:
+        image: 'amazon/dynamodb-local:latest'
+        ports:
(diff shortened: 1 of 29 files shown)


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,177 @@
...
+
+      - name: Setup Redis
+        uses: shogo82148/actions-setup-redis@v1
+        with:
+          redis-version: '7.x'
+          auto-start: 'false'
+
+      - name: Run Redis
+        run: |
+          redis-server --tls-port 6380 --port 6379 \
```
gaby, line 132: All the other DB's are not using TLS, we should remove it for the Benchmark of Redis and Ruedis.

TLS is known to cause performance issues for Redis in terms of Op/sec

gaby (changes requested):

.github/workflows/benchmark.yml
gaby: Seems the TestStore doesnt use TLS, so we dont need to do all this setup for the Benchmarks

gaby (approved):

gaby:
May be good to wait until #993 is merged, but overall LGTM 💪

ReneWerner87 (approved):

PR #1047, opened 2023-10-24
New Storage Driver: ScyllaDb

Chris-Basson-Grundling:
Hey All.

This pr aims to add new Storage implementation of ScyllaDb.

Thanks for the reviews.


.github/workflows/test-scylladb.yml
```diff
@@ -0,0 +1,39 @@
...
+
+jobs:
+  Tests:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        go-version:
+          - 1.19.x
+          - 1.20.x
```
gaby, line 22: Add 1.21.x here

scylladb/README.md
```diff
@@ -0,0 +1,136 @@
+---
+id: scylladb
+title: ScyllaDb
+---
+
+# ScyllaDb
+
+A ScyllaDb storage engine for [Fiber](github.com/gofiber/fiber) using [gocql](github.com/gocql/gocql).
```
gaby, line 8: Is this the right lib? Gocql?

scylladb/README.md
```diff
@@ -0,0 +1,136 @@
...
+func (s *Storage) Get(key string) ([]byte, error)
+func (s *Storage) Set(key string, value []byte, expire time.Duration) error
+func (s *Storage) Delete(key string) error
+func (s *Storage) Reset() error
+func (s *Storage) Close() error
+func (s *Storage) Conn() *gocql.Session
+```
+
+### Installation
+ScyllaDb is tested on the 2 last [Go versions](https://golang.org/dl/) with support for modules. So make sure to initialize one first if you didn't do that yet:
```
gaby, line 29: If so remove 1.19.x from ci file

scylladb/go.mod
```diff
@@ -0,0 +1,14 @@
+module github.com/gofiber/storage/scylladb
+
+go 1.19
```
gaby, line 3: Use 1.20.x here

scylladb/scylladb_test.go
```diff
@@ -0,0 +1,114 @@
+package scylladb
+
+import (
+	"github.com/gofiber/utils"
```
gaby, line 4: All the storage tests should use testify for Asserts. See here https://github.com/gofiber/storage/blob/main/redis/redis_test.go

gaby (changes requested):

scylladb/README.md
gaby: According to their docs we should be using this https://github.com/scylladb/gocql

Source: https://opensource.docs.scylladb.com/stable/using-scylla/drivers/cql-drivers/scylla-go-driver.html

ReneWerner87:
Pls add a section in the root README.md

scylladb/config.go
```diff
@@ -0,0 +1,92 @@
+package scylladb
+
+import "github.com/gocql/gocql"
+
+type Config struct {
```
gaby, line 5: One thing that I see missing is support for setting a TLS config.

scylladb/config.go
```diff
@@ -0,0 +1,92 @@
...
+	//
+	// Optional. Default is nil
+	Session *gocql.Session
+
+	// Keyspace name
+	//
+	// Optional. Default is "fiber"
+	Keyspace string
+
+	// Host name where the ScyllaDb cluster is hosted
```
gaby, line 16: Update the comment to indicate it supports multiple hosts

scylladb/go.mod
```diff
@@ -0,0 +1,19 @@
+module github.com/gofiber/storage/scylladb
+
+go 1.20
+
+require (
+	github.com/gocql/gocql v1.6.0
+	github.com/gofiber/utils v1.1.0
```
gaby, line 7: Running `go mod tidy` should get rid of `gofiber/utils` 🤔 i dont see it being usef anywhere, right?

scylladb/config.go
Chris-Basson-Grundling: Added support for setting TLS config.

scylladb/go.mod
```diff
@@ -0,0 +1,18 @@
...
+
+require (
+	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/golang/snappy v0.0.3 // indirect
+	github.com/hailocab/go-hostpool v0.0.0-20160125115350-e80d13ce29ed // indirect
+	github.com/pmezard/go-difflib v1.0.0 // indirect
+	gopkg.in/inf.v0 v0.9.1 // indirect
+)
+
+replace github.com/gocql/gocql => github.com/scylladb/gocql v1.11.1
```
gaby, line 18: I dont know if dependabot will track this, may be better to fix the require import at the top of the go.mod

scylladb/scylladb_test.go
```diff
@@ -0,0 +1,151 @@
...
+	var (
+		key = "john"
+		val = []byte("doe")
+		exp = 1 * time.Second
+	)
+
+	err := testStore.Set(key, val, exp)
+	require.NoError(t, err)
+
+	time.Sleep(1100 * time.Millisecond)
```
gaby, line 58: There needs to be a `get()` after the sleep, to verify that the `key` expired after 1sec

gaby (changes requested):
Overall looks good, one small change to the tests left

scylladb/go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/gofiber/storage/scylladb
 
 go 1.20
 
+replace github.com/gocql/gocql => github.com/scylladb/gocql v1.11.1
```
gaby, line 5: I mean the import on line 8 should be the one in the replace. I'm not sure if dependabot will pick it up as a "replace" directive

gaby (changes requested):

scylladb/scylladb.go
```diff
@@ -137,7 +137,8 @@ func (s *Storage) Get(key string) ([]byte, error) {
 		}
 		return nil, err
 	}
-	if expiration > 0 && time.Now().Unix() > expiration {
+	// If the expiration time has already passed, then return nil
+	if expiration != 0 && expiration <= time.Now().Unix() {
```
gaby, line 141: Does this mean that the entry stays in the DB? Maybe we should call delete instead once it's expired

gaby (changes requested):

scylladb/scylladb.go
Chris-Basson-Grundling: Great Point, not sure how I missed that, will update shortly.

scylladb/go.mod
Chris-Basson-Grundling: From the tests it should be fine.
https://github.com/dependabot/dependabot-core/blob/main/go_modules/spec/fixtures/go_mods/go.mod

scylladb/scylladb.go
Chris-Basson-Grundling: Updated to use default scylladb TTL

scylladb/scylladb.go
```diff
@@ -0,0 +1,170 @@
...
+// New creates a new storage
+func New(config ...Config) *Storage {
+	var err error
+	var session *gocql.Session
+
+	// Set default config
+	cfg := configDefault(config...)
+
+	if cfg.Keyspace == "" {
+		panic(keyspaceMsg)
```
gaby, line 40: I don't think this ever happens, the configDefault() call above will set the Keyspace to the default value if empty. One thing that could happen is if the user set the Keyspace to something like "   ".

scylladb/config.go
```diff
@@ -0,0 +1,101 @@
...
+	if cfg.Port <= 0 {
+		cfg.Port = ConfigDefault.Port
+	}
+	if cfg.Table == "" {
+		cfg.Table = ConfigDefault.Table
+	}
+	if cfg.Keyspace == "" {
+		cfg.Keyspace = ConfigDefault.Keyspace
+	}
+	if cfg.Consistency == "" {
```
gaby, line 97: You can use:

```go
if len(strings.TrimSpace(yourVariable)) == 0 {
  // set default here
}
```

I just realized this needs to be fixed across all drivers. That's a different issue to deal with later. :-)

scylladb/scylladb.go
gaby: Added a comment above to fix that issue in configDefault()

scylladb/scylladb.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	if cfg.Session == nil {
+		// Create a new cluster
+		cluster := gocql.NewCluster(cfg.Hosts...)
+		cluster.Consistency = gocql.ParseConsistency(cfg.Consistency)
+		cluster.Port = cfg.Port
+
+		// Set credentials if provided
+		if cfg.Username != "" && cfg.Password != "" {
```
gaby, line 50: This onen needs to be updated too

gaby (changes requested):
Looks good overall and tests are passing. Just a small tweak leff to do.

.github/workflows/benchmark.yml
```diff
@@ -117,6 +117,11 @@ jobs:
         run: |
           docker run -d --restart always -p 9000:9000 --name storage-minio -e MINIO_ROOT_USER='minio-user' -e MINIO_ROOT_PASSWORD='minio-password' minio/minio server /data
 
+      - name: Install ScyllaDb
+        run: |
+          docker run --name scylladb -p 9042:9042 -p 19042:19042 -p 9160:9160 -p 7000:7000 -p 7001:7001 -p 7199:7199 -p 9180:9180 -d scylladb/scylla:latest --broadcast-address 127.0.0.1 --listen-address 0.0.0.0 --broadcast-rpc-address 127.0.0.1
+          sleep 30 # Wait for ScyllaDb to initialize
```
efectn, line 123: Can we reduce sleep time? Does 15 work stable?

scylladb/scylladb.go
```diff
@@ -0,0 +1,170 @@
...
+	tableName string
+
+	selectQuery string
+	insertQuery string
+	deleteQuery string
+	resetQuery  string
+}
+
+var (
+	checkSchemaMsg = "the `value` row has an incorrect data type. " +
```
efectn, line 22: Can you create error vars with errors.New

scylladb/scylladb.go
```diff
@@ -0,0 +1,170 @@
...
+
+	// Create keyspace if it does not exist
+	if err = session.Query(fmt.Sprintf(createKeyspaceQuery, cfg.Keyspace)).Exec(); err != nil {
+		session.Close()
+		panic(err)
+	}
+
+	// Drop table if reset is true
+	if cfg.Reset {
+		if err = session.Query(fmt.Sprintf(dropQuery, cfg.Keyspace, cfg.Table)).Exec(); err != nil {
```
efectn, line 80: Just curious. Do we get panic if bucket is empty?

scylladb/scylladb.go
```diff
@@ -0,0 +1,170 @@
...
+			session.Close()
+			panic(err)
+		}
+	}
+
+	// Create the storage
+	store := &Storage{
+		session:     session,
+		tableName:   cfg.Table,
+		selectQuery: fmt.Sprintf("SELECT value FROM %s.%s WHERE key = ?", cfg.Keyspace, cfg.Table),
```
efectn, line 90: Can you make these unexported vars as well?

scylladb/scylladb.go
Chris-Basson-Grundling: Yes, you will get the error message "gocql: no streams available on connection"
Maximum number of inflight requests allowed per connection defaults to 3276.

gaby (approved):
LGTM! Thanks for the PR

ReneWerner87 (approved):

Issue #995, opened 2023-09-18
🤗 [feature]: Add support for Keys() to Redis and Memory drivers


glimchb:
### Question Description

Is there a plan to extend this interface to support fetching all keys? 

### Code Snippet (optional)

```go
package main

import "github.com/gofiber/storage/%package%"

func main() {
  // Steps to reproduce
}
```

### Checklist:

- [X] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [X] I have checked for existing issues that describe my questions prior to opening this one.
- [X] I understand that improperly formatted questions may be closed without explanation.

gaby:
@glimchb This would normally be easy, but each DB has their own way, some don't even support this. For example with redis you can do this

```go
store := redis.New()
keys := store.Conn().Keys()
```

glimchb:
understood @gaby 

so no plans to add this new function to the interface [itself](https://github.com/gofiber/storage/blob/main/storage.go) or it is under consideration ?

gaby:
@glimchb We can easily add it to Redis and Rueidis. Do you have any specific driver in mind that you need?

glimchb:
Thank you @gaby

yes, Redis for production and memory for mocking

gaby:
@glimchb I will work on adding that to redis and rueidis this week

glimchb:
@gaby appreciate this!

gaby:
@glimchb ListAll, GetAll, Keys are all pretty much the same right?

I'm going with `Keys()`, which is the same name that Redis uses.

gaby:
@glimchb PR's have been submitted for both `Redis` and `Memory`

glimchb:
> I'm going with `Keys()`, which is the same name that Redis uses.
excellent!

glimchb:
> @glimchb PR's have been submitted for both `Redis` and `Memory`

one small comment on reds PR #1052

gaby:
Fixed via #1052 #1053 pending Release of new Redis/Memory drivers

gaby:
@glimchb New releases are out, for memory the major version changed to v2.0.0, for Redis it's under v3.1.0

glimchb:
> @glimchb New releases are out, for memory the major version changed to v2.0.0, for Redis it's under v3.1.0

amazing!

glimchb:
[...]

PR #1052, opened 2023-10-30
Add support for Keys() to Redis Driver

gaby:
- Added `Keys()` method using `SCAN` to avoid performance issues with large sets of keys
- Fixed the expiration test which was checking for the key in a separate test
- Removed the global `testStore`, each test now creates a `testStore`. 
- Added calls to `Keys()` to all tests.
- Updated Readme

Partial fix for #995 I will be submitting a different PR for the other drivers requested by the user

### DIFF
--- a/.github/workflows/test-redis.yml
+++ b/.github/workflows/test-redis.yml
@@ -66,6 +66,9 @@ jobs:
                     slave3-port: 7005
                     sleep-duration: 10
 
+            -   name: Wait for Redis to Start
+                run: sleep 15
+
             -   name: Install Go
                 uses: actions/setup-go@v4
                 with:
--- a/redis/redis.go
+++ b/redis/redis.go
@@ -74,8 +74,6 @@ func New(config ...Config) *Storage {
 	}
 }
 
-// ...
-
 // Get value by key
 func (s *Storage) Get(key string) ([]byte, error) {
 	if len(key) <= 0 {
@@ -117,4 +115,33 @@ func (s *Storage) Close() error {
 // Return database client
 func (s *Storage) Conn() redis.UniversalClient {
 	return s.db
+}
+
+// Return all the keys
+func (s *Storage) Keys() ([][]byte, error) {
+	var keys [][]byte
+	var cursor uint64
+
+	for {
+		var batch []string
+		var err error
+		batch, cursor, err = s.db.Scan(context.Background(), cursor, "*", 10).Result()
+		if err != nil {
+			return nil, err
+		}
+
+		for _, key := range batch {
+			keys = append(keys, []byte(key))
+		}
+
+		if cursor == 0 {
+			break
+		}
+	}
+
+	if len(keys) == 0 {
+		return nil, nil
+	}
+
+	return keys, nil
 }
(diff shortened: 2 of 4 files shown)


.github/workflows/test-redis.yml
```diff
@@ -66,6 +66,9 @@ jobs:
                     slave3-port: 7005
                     sleep-duration: 10
 
+            -   name: Wait for Redis to Start
```
gaby, line 69: We may not need this, i'm just tired of the inconsistency of Redis Cluster timing out 😆 So this solves that issue

gaby:
Benchmark is failing because of `coherence` driver: 

```
| Benchmark suite | Current: 0b9f4aebc8b4e7c2d8bbfbfdc50449b319ffbb33 | Previous: ddac78a1dd60fec70e[58](https://github.com/gofiber/storage/actions/runs/6686601876/job/18166161978?pr=1052#step:14:59)e1c641cf589962944cdd | Ratio |
|-|-|-|-|
| `Benchmark_Coherence_Set - ns/op` | `9263314` ns/op | `2005652` ns/op | `4.62` |
```

redis/redis.go
```diff
@@ -118,3 +116,32 @@ func (s *Storage) Close() error {
...
 	return s.db
 }
+
+// Return all the keys
+func (s *Storage) Keys() ([][]byte, error) {
+	var keys [][]byte
+	var cursor uint64
+	var err error
+
+	for {
```
glimchb, line 126: wdyt about using iterator instead like this?
```
iter := rdb.Scan(ctx, 0, "prefix:*", 0).Iterator()
for iter.Next(ctx) {
	fmt.Println("keys", iter.Val())
}
if err := iter.Err(); err != nil {
	panic(err)
}
```

gaby: I could but my implementation is based on Redis official documentation for using `scan()`: 

See here: https://pkg.go.dev/github.com/redis/go-redis/v9#Client.Scan

.github/workflows/test-redis.yml
ReneWerner87: really 15 seconds? can we shrink this a little bit

redis/redis_test.go
```diff
@@ -34,10 +36,17 @@ func Test_Redis_Set_Override(t *testing.T) {
 
 	err = testStore.Set(key, val, 0)
 	require.NoError(t, err)
+
+	keys, err := testStore.Keys()
```
ReneWerner87, line 40: i have seen that you always test the new function in other tests 
but would prefer if we had a separate function for the keys functionality so we don't always bloat all the others

ReneWerner87 (approved):

redis/redis_test.go
gaby: @ReneWerner87 To be able to test `Keys()` there needs to be data. Doing it this way not only tests the Keys() function but also validates that the keys from other actions.

I can write a separate test for `Keys()` but we should still check that Keys() returns 1 in the test for `set()` after creating a key. Or that it returns 0 after deleting the last key().

efectn (approved):

PR #1053, opened 2023-10-30
Add support for Keys() to Memory Driver

gaby:
- Added Keys() method
- Fixed the expiration test which was checking for the key in a separate test
- Removed the global testStore, each test now creates a testStore.
- Added calls to Keys() to all tests.
- Updated Readme

Second Part fix for https://github.com/gofiber/storage/issues/995
First part #1052

### DIFF
--- a/memory/memory_test.go
+++ b/memory/memory_test.go
@@ -81,6 +81,37 @@ func Test_Storage_Memory_Set_Expiration(t *testing.T) {
 	require.Nil(t, keys)
 }
 
+func Test_Storage_Memory_Set_Long_Expiration_with_Keys(t *testing.T) {
+	var (
+		testStore = New()
+		key       = "john"
+		val       = []byte("doe")
+		exp       = 5 * time.Second
+	)
+
+	keys, err := testStore.Keys()
+	require.NoError(t, err)
+	require.Nil(t, keys)
+
+	err = testStore.Set(key, val, exp)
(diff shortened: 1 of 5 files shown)


gaby:
Did I find a bug in the Memory storage? Why are there values in `Keys` if they have been deleted/expired?

gaby:
Seems like `Get()` is faking the missing key, and the GC hasn't run yet. So when using `Keys()` the values are actually still in memory.

gaby:
Done, added expiration check to `Keys()` and a new unit-test to call `Keys()` during a long expiration.

ReneWerner87 (approved):

memory/memory_test.go
```diff
@@ -44,53 +53,104 @@ func Test_Storage_Memory_Get(t *testing.T) {
...
+	keys, err = testStore.Keys()
+	require.NoError(t, err)
+	require.Len(t, keys, 1)
+
+	time.Sleep(4000 * time.Millisecond)
 	result, err := testStore.Get(key)
 	require.NoError(t, err)
 	require.Zero(t, len(result))
+
+	keys, err = testStore.Keys()
```
ReneWerner87, line 110: i have seen that you always test the new function in other tests 
but would prefer if we had a separate function for the keys functionality so we don't always bloat all the others

efectn (approved):

PR #1087, opened 2023-11-15
Support for dynamic golangci-lint matrix

gaby:
- Fixes to dynamically generate a matrix for testing with golangci-lint
- If we run `--enable-all` there's several failures that should be address in a different PR


gaby:
I will fix the ones that show up without `--enable-all` tonight.

ReneWerner87 (approved):

ReneWerner87 (approved):

PR #1193, opened 2024-01-29
New storage driver: NATS

cr1cr1:
Ref: https://docs.nats.io/nats-concepts/jetstream/key-value-store


.github/workflows/test-nats.yml
```diff
@@ -0,0 +1,31 @@
...
+    push:
+        branches:
+            - master
+            - main
+        paths:
+            - 'nats/**'
+    pull_request:
+        paths:
+            - 'nats/**'
+name: "Tests Local Storage"
```
gaby, line 11: Test NATS Driver

nats/README.md
```diff
@@ -0,0 +1,122 @@
...
+---
+
+
+![Release](https://img.shields.io/github/v/tag/gofiber/storage?filter=nats*)
+[![Discord](https://img.shields.io/discord/704680098577514527?style=flat&label=%F0%9F%92%AC%20discord&color=00ACD7)](https://gofiber.io/discord)
+![Test](https://img.shields.io/github/actions/workflow/status/gofiber/storage/test-nats.yml?label=Tests)
+![Security](https://img.shields.io/github/actions/workflow/status/gofiber/storage/gosec.yml?label=Security)
+![Linter](https://img.shields.io/github/actions/workflow/status/gofiber/storage/linter.yml?label=Linter)
+
+An NATS storage driver.
```
gaby, line 13: A NATS Key/Value Storage Driver.

nats/config.go
```diff
@@ -0,0 +1,97 @@
...
+	MaxReconnects int
+	// Nats context
+	Context context.Context
+	// Nats key value config
+	KeyValueConfig jetstream.KeyValueConfig
+	Logger         *slog.Logger
+	// Applicable only if Logger is nil.
+	// Until go 1.22, it is weird to set log level.
+	// See https://github.com/golang/go/issues/62418
+	LogLevel slog.Level
```
gaby, line 36: Can we do this without slog? So that we can support 1.20?

nats/config.go
```diff
@@ -0,0 +1,97 @@
...
+						Level: cfg.LogLevel,
+					},
+				),
+			)
+		} else {
+			cfg.Logger = ConfigDefault.Logger
+		}
+	}
+	// if !cfg.RetryOnFailedConnect {
+	// 	cfg.RetryOnFailedConnect = ConfigDefault.RetryOnFailedConnect
```
gaby, line 93: ???

nats/README.md
```diff
@@ -0,0 +1,122 @@
...
+// Initialize custom config
+store := nats.New(nats.Config{
+ GCInterval: 10 * time.Second,
+})
+```
+
+### Config
+
+```go
+type Config struct {
```
gaby, line 75: Can we add support for TLSContext?

nats/config.go
cr1cr1: You are right, I should not force go upgrade just because of logging. I just found the standard log ..lacking.

nats/README.md
cr1cr1: Yes! working on it

nats/config.go
cr1cr1: Umm... could not decide whether start the driver in "disconnected" mode, and have automatic retries later, because I do not feel comfortable for this driver to panic if it does not establish a connection, rather gracefully fall back to in-memory.

What do you think about that?

gaby: @cr1cr1 All the other drivers panic if they can't connect during Startup. See Redis for example.

gaby: So yes, it should auto start/connect and panic if it can't

cr1cr1: Yes, I have seen they panic. By default, this driver will attempt to connect using defaults then give up and panic.

gaby: Ok, you can add `RetryOnFailedConnect` just make the default be "False"

nats/config.go
cr1cr1: What do you say about using something like zerolog instead?

nats/config.go
cr1cr1: That's what it is now

gaby: Perfect

nats/config.go
gaby: Either Fiber Logger or Fiber Zerolog from contrib: https://github.com/gofiber/contrib/tree/main/fiberzerolog

gaby:
@cr1cr1 After line 40 of .github/workflows/gosec.yml add this:

```yaml
dir_names_max_depth: '1'
```

nats/README.md
```diff
@@ -0,0 +1,113 @@
...
+---
+
+
+![Release](https://img.shields.io/github/v/tag/gofiber/storage?filter=nats*)
+[![Discord](https://img.shields.io/discord/704680098577514527?style=flat&label=%F0%9F%92%AC%20discord&color=00ACD7)](https://gofiber.io/discord)
+![Test](https://img.shields.io/github/actions/workflow/status/gofiber/storage/test-nats.yml?label=Tests)
+![Security](https://img.shields.io/github/actions/workflow/status/gofiber/storage/gosec.yml?label=Security)
+![Linter](https://img.shields.io/github/actions/workflow/status/gofiber/storage/linter.yml?label=Linter)
+
+An NATS Key/Value storage driver.
```
gaby, line 13: I think this is "A" instead of "An" ?

nats/nats.go
```diff
@@ -0,0 +1,368 @@
...
+// logErrorw is a helper function to log error messages
+func (s *Storage) logErrorw(msg string, keysAndValues ...interface{}) {
+	if s.cfg.UseLogger && s.cfg.Logger != nil {
+		s.cfg.Logger.Errorw(msg, keysAndValues...)
+	}
+}
+
+// logInfow is a helper function to log error messages
+func (s *Storage) logInfow(msg string, keysAndValues ...interface{}) {
+	if s.cfg.UseLogger && s.cfg.Logger != nil {
```
gaby, line 46: Do we need the double check here? Isnt checking for nil enough?

nats/nats.go
```diff
@@ -0,0 +1,368 @@
...
+	Expiry int64
+}
+
+func init() {
+	gob.Register(entry{})
+}
+
+// logErrorw is a helper function to log error messages
+func (s *Storage) logErrorw(msg string, keysAndValues ...interface{}) {
+	if s.cfg.UseLogger && s.cfg.Logger != nil {
```
gaby, line 39: Same here

nats/nats.go
gaby: Nvmd i see it now

nats/config.go
```diff
@@ -0,0 +1,71 @@
...
+	// Nats connection name
+	ClientName string
+	// Nats context
+	Context context.Context
+	// Nats key value config
+	KeyValueConfig jetstream.KeyValueConfig
+	// Logger. Using Fiber provides the AllLogger interface for adapting the various log libraries.
+	Logger log.AllLogger
+	// Use the Logger for nats events, default: false
+	UseLogger bool
```
gaby, line 26: Rename this to something like `Verbose`

nats/nats.go
```diff
@@ -0,0 +1,368 @@
...
+	)
+	if err != nil {
+		s.err = append(s.err, fmt.Errorf("subject %q: %w", sub.Subject, err))
+	}
+}
+
+// closedHandler is a helper function to set the closed handler
+func (s *Storage) closedHandler(nc *nats.Conn) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
```
gaby, line 117: Why get a lock here to log a message? This will reduce performance with logs enable.

nats/nats.go
```diff
@@ -0,0 +1,368 @@
...
+	if opErr, ok := err.(*net.OpError); ok && natsOpts.RetryOnFailedConnect {
+		if opErr.Op != "dial" {
+			panic(err)
+		}
+	} else if err != nil {
+		panic(err)
+	}
+
+	// TODO improve this crude way to wait for the connection to be established
+	time.Sleep(100 * time.Millisecond)
```
gaby, line 201: Maybe make this a config value. Could be named `ConnectTimeout` with a default val of 100ms

nats/nats.go
```diff
@@ -36,14 +36,14 @@ func init() {
 
 // logErrorw is a helper function to log error messages
 func (s *Storage) logErrorw(msg string, keysAndValues ...interface{}) {
-	if s.cfg.UseLogger && s.cfg.Logger != nil {
+	if s.cfg.Logger != nil {
```
gaby, line 39: I actuallt think we need the check 😂 but just the first one. This since the var will have a default value of either True or False. It won't be nil at any point

gaby: so `if s.cfg.UseLogger { ... }` should be enough

gaby: @cr1cr1 Does the above make sense?

cr1cr1: Indeed, Logger is never nil when using the bool option, renamed to Verbose

nats/nats.go
cr1cr1: It is not technically a ConnectTimeout, just a delay to allow the async handlers to get a connection and create the KV instance.

gaby:
@cr1cr1 Thank you, will do a full review within the next day.

nats/nats.go
gaby: Yup!

nats/testdata/certs/ca.pem
```diff
@@ -0,0 +1,27 @@
+-----BEGIN CERTIFICATE-----
```
ReneWerner87, line 1: is it really needed ? we have this step for the other storages https://github.com/gofiber/storage/blob/1e017301cd303f880c93c658ad67d83b9c848294/.github/workflows/test-rueidis.yml#L27-L28

cr1cr1: Sorry, I just merged some recent changes. rueidis is not my work :)

ReneWerner87: sorry was just dragged into another topic

yes, instead of checking in fixed certificates, we could do it like the other adapters and generate them on the fly

README.md
```diff
@@ -71,4 +71,4 @@ type Storage interface {
 - [S3](./s3/README.md) <a href="https://github.com/gofiber/storage/actions?query=workflow%3A%22Tests+S3%22"> <img src="https://img.shields.io/github/actions/workflow/status/gofiber/storage/test-s3.yml?branch=main&label=%F0%9F%A7%AA%20&style=flat&color=75C46B" /> </a>
 - [ScyllaDB](./scylladb/README.md) <a href="https://github.com/gofiber/storage/actions?query=workflow%3A%22Tests+scylladb%22"> <img src="https://img.shields.io/github/actions/workflow/status/gofiber/storage/test-scylladb.yml?branch=main&label=%F0%9F%A7%AA%20&style=flat&color=75C46B" /> </a>
 - [SQLite3](./sqlite3/README.md) <a href="https://github.com/gofiber/storage/actions?query=workflow%3A%22Tests+Sqlite3%22"> <img src="https://img.shields.io/github/actions/workflow/status/gofiber/storage/test-sqlite3.yml?branch=main&label=%F0%9F%A7%AA%20&style=flat&color=75C46B" /> </a>
-
+- [NATS Key/Value Store](./nats/README.md) <a href="https://github.com/gofiber/storage/actions?query=workflow%3A%22Tests%20Nats%20Driver%22"> <img src="https://img.shields.io/github/actions/workflow/status/gofiber/storage/test-nats.yml?branch=main&label=%F0%9F%A7%AA%20&style=flat&color=75C46B" /> </a>
```
ReneWerner87, line 74: @cr1cr1 
please insert in alphabetical order above
and remove the suffix "Key/Value store" -> they are all key/value stores

nats/testdata/certs/ca.pem
ReneWerner87: @cr1cr1 can you adjust this ?

cr1cr1: Yes, working on it, but it is painful to test github actions locally.

ReneWerner87: have you tried [act](https://github.com/nektos/act) 
with this i can use the workflow locally

f.e.
image

cr1cr1: Yes, that's what I'm using but the generate files by the gen-test-certs.sh are not mounted in the nats-server docker using **act** because it seems they are generated in another container.

ReneWerner87 (approved):

gaby:
> LGTM @gaby also ok ?

Looking at it now, i'm not convinced its using TLS. Let me check the docs

nats/nats_test.go
```diff
@@ -0,0 +1,269 @@
...
+	"github.com/nats-io/nats.go/jetstream"
+	"github.com/stretchr/testify/require"
+)
+
+var config = Config{
+	URLs: "nats://localhost:4443",
+	NatsOptions: []nats.Option{
+		nats.MaxReconnects(2),
+		// Enable TLS by specifying RootCAs
+		nats.RootCAs("../tls/ca.crt"),
```
gaby, line 17: This is missing the following:

```go
nats.ClientCert("client-cert.pem", "client-key.pem")
```

Source https://docs.nats.io/using-nats/developer/connecting/tls#connecting-with-tls-and-verify-client-identity

nats/testdata/nats-tls.conf
```diff
@@ -0,0 +1,10 @@
+# Simple TLS config file
+
+port: 4443
+net: 0.0.0.0 # net interface
+
+tls {
+  cert_file:  "/tls/redis.crt"
+  key_file:   "/tls/redis.key"
```
gaby, line 8: `ca_file` is missing here: https://docs.nats.io/running-a-nats-service/configuration/securing_nats/tls

gaby: During the CI, the file will be `/tls/ca.crt`

nats/nats_test.go
gaby: These are the generated paths during CI:

- /home/runner/work/storage/storage/tls/client.crt
- /home/runner/work/storage/storage/tls/client.key

gaby (changes requested):

nats/nats_test.go
cr1cr1: Just by specifying `nats.RootCAs()`, will use TLS connection, but will be an anonymous one. This is the minimum.

Indeed, a good idea to test cert auth as well. Will implement.

gaby (approved):
LGTM

ReneWerner87:
https://github.com/gofiber/storage/releases/tag/nats%2Fv1.0.0

ReneWerner87:
https://docs.gofiber.io/storage/nats_v1.x.x/nats/

PR #1335, opened 2024-04-06
Add DeleteMany() to S3 Storage

xEricL:
The aws s3 driver has a nice `DeleteObjects()` method for bulk-deleting multiple objects in a bucket. I added a basic `DeleteMany(keys ...string)` method that allows users to provide multiple keys to delete a set of objects in a single request.


gaby (approved):
LGTM

PR #1530, opened 2024-11-13
Add MaxRetry for MinIO operations

itpey:
This PR introduces a new MaxRetry configuration option to the Config struct, allowing users to specify the maximum number of retry attempts for MinIO operations.

### DIFF
--- a/minio/README.md
+++ b/minio/README.md
@@ -94,6 +94,10 @@ type Config struct {
     // Optional. Default is false
     Reset bool
     
+    // The maximum number of times requests that encounter retryable failures should be attempted.
+	// Optional. Default is 10, same as the MinIO client.
+	MaxRetry int
+
     // Credentials Minio access key and Minio secret key.
     // Need to be defined
     Credentials Credentials
@@ -124,6 +128,7 @@ var ConfigDefault = Config{
     Token:               "",
     Secure:              false,
     Reset:               false,
+
     Credentials:         Credentials{},
     GetObjectOptions:    minio.GetObjectOptions{},
     PutObjectOptions:    minio.PutObjectOptions{},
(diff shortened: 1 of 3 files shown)


ReneWerner87:
Pls add a test case

itpey:
> Thx for the pr
> 
> Pls add a test case


I don't think a new test case is necessary since this feature is already covered and well-tested within the MINIO client itself. I have not implemented a new retry mechanism.

https://github.com/minio/minio-go/blob/ab1c2fe050a4dd502c863178696a54c44610ca82/retry.go

minio/README.md
```diff
@@ -94,6 +94,10 @@ type Config struct {
     // Optional. Default is false
     Reset bool
     
+    // The maximum number of times requests that encounter retryable failures should be attempted.
+	// Optional. Default is 10, same as the MinIO client.
+	MaxRetry int
```
ReneWerner87, line 99:
```suggestion
    // Optional. Default is 10, same as the MinIO client.
    MaxRetry int
```

ReneWerner87:
> > Thx for the pr
> > Pls add a test case
> 
> I don't think a new test case is necessary since this feature is already covered and well-tested within the MINIO client itself. I have not implemented a new retry mechanism.
> 
> [minio/minio-go@`ab1c2fe`/retry.go](https://github.com/minio/minio-go/blob/ab1c2fe050a4dd502c863178696a54c44610ca82/retry.go)

this should not be used to check the underlying functionality, but only to check that the configuration has been set correctly

this prevents an error from occurring later if the configurations become much larger

ReneWerner87:
but it looks like I'm a bit too strict there
at least we don't have that with the other adapters
only with the middleware
maybe we can add that later
would definitely help to find configuration errors early on when this process becomes more complex

ReneWerner87 (approved):

PR #1552, opened 2025-01-16
fix(minio): close object after GetObject has returned

NuttyShrimp:
This prevents a goroutine leak by the minio client from happening.
I found this issue because a data retrieval endpoint kept it's minio goroutine alive.
Then I stumbled upon https://github.com/minio/minio-go/issues/1508 and saw the fiber wrapper didn't close the object!

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved resource management when retrieving objects from MinIO storage by ensuring proper object closure.

### DIFF
--- a/minio/minio.go
+++ b/minio/minio.go
@@ -75,6 +75,7 @@ func (s *Storage) Get(key string) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


gaby (approved):
👍 LGTM

PR #1536, opened 2024-12-28
chore(ci): run benchmarks for each storage in separate workers

mdelapenya:
## What does this PR do?
It adds an initial job in the benchmark workflow in order to identify which modules where modified for a given build.

To detect the changes, we have created a shell script, `./github/scripts/changed-modules.sh`. This script receives one single env var as input parameter, `ALL_CHANGED_FILES`, which on CI will be provided by a Github action, https://github.com/tj-actions/changed-files, that puts in there a list of the modified files, comparing the current PR changeset with the parent branch (main). We can tune this up to always check the latest `main` branch (open to discussion here).

The script will compare all existing modules (looking up all the `go.mod` files of interest, no test files) with the modified files, building an array of modified modules. At the moment there is a modified file in the .github directory, all the modules will be included in the build.

Finally, for running the benchmarks, we are adding a matrix for running them for the modified storage module, with the following rules:
- all modules will run the benchmarks on `ubuntu-latest`.
- services that are needed by a module will be started conditionally on the CI (e.g. redis, etcd)
- the output file will prepend the module name, so that the final save-benchmark step can compare against that module, alone.

## Why is this important?
Separate concerns and better troubleshoot of issues with benchmarks: at the moment, all benchmarks are executed in serie, which 1) slows down the build, 2) could cause a PR to fail if one benchmark for another module failed.

With these changes we try to improve the contribution experience when working on one module.

PR #1553, opened 2025-01-17
🧹 chore: Initial simplification of workflows

gaby:
- Remove `govulncheck` workflow, it adds no value.
- Remove `gosec` workflow, it adds no value.
- Remove `codeql` workflow, it adds no value.
- Simplify the `release-drafter` workflows into one using `dorny/paths-filter`

After this we can look into simplifying the benchmark workflow to faciliate #1536


ReneWerner87:
nice idea

mdelapenya:
> Simplify the release-drafter workflows into one using dorny/paths-filter

Cool! It would simplify my PR even more, thanks for letting me know, as I did not know about that GH action

gaby:
> > Simplify the release-drafter workflows into one using dorny/paths-filter
> 
> Cool! It would simplify my PR even more, thanks for letting me know, as I did not know about that GH action

Yes, it even has support for `matrix`. I just committed another change to use that

ReneWerner87:
> > Simplify the release-drafter workflows into one using dorny/paths-filter
> 
> Cool! It would simplify my PR even more, thanks for letting me know, as I did not know about that GH action

we also want to reduce the configs and try to create the config on the fly with a template by replacing the folder name in this template 

and maybe we can make it so that we don't have to define the individual storage either 

then we can also apply the whole thing to template and contrib later on

ReneWerner87 (approved):

PR #1562, opened 2025-01-30
🔥 feat: Add support for Neo4j

### LINKED ISSUE #1561: 🤔 [Question]: Is there a particular reason why Neo4j isn't included among the storage drivers?
### Question Description

I'm just surprised that even with it's wide adoption in the graph database ecosystem, a storage driver isn't available for it in gofiber.

I started working on one&#x2014;since I've search everywhere and no one seems to be talking about it. But then I thought I should ask first, perhaps there might be a reason why it's not supported.

### Code Snippet (optional)

```go
package main

import (
  "time"
  
  neo4j "github.com/neo4j/neo4j-go-driver/v5/neo4j"
  neo4jconfig "github.com/neo4j/neo4j-go-driver/v5/neo4j/config"

  neo4jstore "github.com/gofiber/storage/neo4j"
)

type Config struct {
  // Optional. Defaults to nil. Overrides the following connection parameters
  Driver neo4j.DriverWithContext

  TargetBoltURI string

  // Optional. Defaults to basic authentication with the provided username/password and "" realm value
  AuthToken neo4j.AuthToken

  // Optional. Defaults to []func(*neo4jconfig.Config){}
  Configurers []func(*neo4jconfig.Config)

  Username string

  Password string

  // If you have the administrative privilege, else it will be ignored for the default database, "neo4j"
  // Optional. Default is "fiber" (still if you have the administrative privilege)
  Database string

  // Optional. Default is "fiber_storage"
  Node string

  // Optional. Default is false
  Reset bool

  GCInterval time.Duration
}

func main() {
  // Initialize default config
  store := neo4jstrg.New()

  // Initialize custom config
  store := neo4jstore.New(neo4jstore.Config{
    BoltURI:       "neo4j://localhost:7687",
    Database:      "fiber",
    Node:          "fiber_storage",
    Reset:         false,
    GCInterval:    10 * time.Second,
  })
}
```

### Checklist:

- [x] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [x] I have checked for existing issues that describe my questions prior to opening this one.
- [x] I understand that improperly formatted questions may be closed without explanation.

the-real-i9:
Neo4j storage driver: Configuration declaration (config.go), Storage implementation (neo4j.go), Tests (neo4j_test.go), and a README.md

arangodb/README.md: typo fix: I think the author meant arangodb, but mistakenly typed mysql

postgres/config.go: removed duplicate `cfg.Table == ""` condition check

fixes #1561 

## Summary by CodeRabbit

- **Documentation**
	- Updated ArangoDB README to correct implementation reference.
	- Added comprehensive README for Neo4j storage driver, detailing functionality, installation, and configuration.
	- Added Neo4j to the list of storage implementations in the main README.

- **New Features**
	- Introduced Neo4j storage driver with key-value store functionality.
	- Added configuration management for Neo4j database connections.
	- Implemented storage operations such as Get, Set, Delete, and Reset.

- **Bug Fixes**
	- Streamlined logic in PostgreSQL configuration management by removing redundant checks.

- **Tests**
	- Created extensive test suite for Neo4j storage driver, including unit and benchmark tests for performance evaluation.


neo4j/neo4j.go
```diff
@@ -0,0 +1,205 @@
...
+	res, err := neo4j.ExecuteQuery(
+		ctx, s.db, s.cypherMatch,
+		map[string]any{"key": key},
+		neo4j.EagerResultTransformer,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	data, _, _ := neo4j.GetRecordValue[map[string]any](res.Records[0], "data")
```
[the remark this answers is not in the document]
the-real-i9: I don't think this will be a problem. I'm using `OPTIONAL MATCH`, `data` will either be `null` if no records are returned.

This condition check follows:
```go
if data == nil {
  return nil, nil
}
```

ReneWerner87:
thx, don´t forget the test setup/workflow

neo4j/README.md
```diff
@@ -0,0 +1,146 @@
...
+}
+```
+
+#### A note on Authentication
+
+If auth is enabled on your server, then authentication must be provided in one of the three ways (the previous overrides the next):
+
+- Via the connection pool, `neo4j.DriverWithContext`, provided on the `DB` field.
+- Via the `Auth` field: it must be an `auth.TokenManager` whose value is any one but `neo4j.NoAuth()`.
+- By setting both `Username` and `Password` fields: This will cause the this storage driver to use Basic Auth.
```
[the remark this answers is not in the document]
the-real-i9: I fixed this

neo4j/neo4j.go
```diff
@@ -0,0 +1,207 @@
...
+	_, _ = neo4j.ExecuteQuery(context.Background(), s.db, s.cypherGC, map[string]any{"exp": t.Unix()}, neo4j.EagerResultTransformer)
+}
+
+func mapToStruct(src map[string]any, dest any) {
+	bt, _ := json.Marshal(src)
+
+	if err := json.Unmarshal(bt, dest); err != nil {
+		panic(err)
+	}
+}
```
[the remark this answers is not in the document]
the-real-i9: It's just a helper function used in the library it's not expected to even result in an error

neo4j/neo4j.go
```diff
@@ -0,0 +1,207 @@
...
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Unable to create connection pool: %v\n", err)
+		}
+	}
+
+	ctx := context.Background()
+
+	if err := db.VerifyConnectivity(ctx); err != nil {
+		panic(err)
+	}
```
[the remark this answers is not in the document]
the-real-i9: This error is not expected to occur

neo4j/README.md
```diff
@@ -0,0 +1,146 @@
+---
+id: neo4j
+title: Neo4j
+---
+
+A Neo4j storage driver using [neo4j/neo4j-go-driver](https://github.com/neo4j/neo4j-go-driver).
+
+> **Note: Requires Go 1.20 and above**
```
gaby, line 8: Should say "Latest two release of Golang"

neo4j/go.mod
```diff
@@ -0,0 +1,16 @@
+module github.com/gofiber/storage/neo4j/v1
+
+go 1.23.1
```
gaby, line 3: Make this `1.23`

neo4j/config.go
```diff
@@ -0,0 +1,102 @@
...
+	//
+	// DB neo4j.DriverWithContext object will override connection uri and other connection fields.
+	//
+	// Optional. Default is nil.
+	DB neo4j.DriverWithContext
+
+	// Target Server
+	//
+	// Optional. Default is "neo4j://localhost"
+	TargetBoltURI string
```
gaby, line 22: Rename to `URI`

neo4j/config.go
```diff
@@ -0,0 +1,102 @@
...
+	//
+	// Auth auth.TokenManager will override Username and Password fields.
+	//
+	// Optional. Default is nil.
+	Auth auth.TokenManager
+
+	// Connection configurations
+	//
+	// Optional. Default is nil
+	Configurers []func(*config.Config)
```
gaby, line 34: Rename to `Configs` or `Configurations`

neo4j/neo4j.go
```diff
@@ -0,0 +1,207 @@
+package neo4jstore
```
gaby, line 1: This should be `neo4j`

neo4j/neo4j.go
```diff
@@ -0,0 +1,207 @@
...
+// New creates a new storage
+func New(config ...Config) *Storage {
+	// Set default config
+	cfg := configDefault(config...)
+
+	// Select db connection
+	var err error
+	db := cfg.DB
+	if db == nil {
+		db, err = neo4j.NewDriverWithContext(cfg.TargetBoltURI, cfg.Auth, cfg.Configurers...)
```
gaby, line 42: Use named params instead of positional

neo4j/neo4j_test.go
```diff
@@ -0,0 +1,175 @@
+package neo4jstore
+
+import (
+	"os"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
```
gaby, line 10: If neo4j has a container, use `testcontainers-go` for testing

gaby (changes requested):

.github/workflows/test-neo4j.yml
```diff
@@ -0,0 +1,40 @@
...
+            neo4j:
+                image: 'neo4j:5.26.1'
+                env:
+                    NEO4J_AUTH: 'neo4j/pass_w_rd'
+                ports:
+                    - '7687:7687'
+        strategy:
+            matrix:
+                go-version:
+                    - 1.20.x
```
gaby, line 25: Remove go 1.20->1.22

.github/workflows/test-neo4j.yml
```diff
@@ -0,0 +1,40 @@
...
+    pull_request:
+        paths:
+            - 'neo4j/**'
+name: "Tests Neo4j"
+jobs:
+    Tests:
+        runs-on: ubuntu-latest
+        services:
+            neo4j:
+                image: 'neo4j:5.26.1'
```
gaby, line 17: Use minor release `5.26`, patch release is too granular.

.github/workflows/test-neo4j.yml
gaby: The workflow wont even work without this change since go.mod points to 1.23

neo4j/neo4j.go
the-real-i9: The package name for `neo4j/neo4j-go-driver` is itself `neo4j`. Should I add a custom name for the neo4j driver package instead.

gaby: In that case, leave it as neo4jstore or neo4jdriver

.github/workflows/test-neo4j.yml
the-real-i9: I'm already working on using the `testcontainers-go` for neo4j. That means I don't need to setup a neo4j service in the workflow for testing purposes.

.github/workflows/test-neo4j.yml
the-real-i9: Yes, I understand now.

I'm working on all changes suggested and will push them at once.

the-real-i9: Yes, I understand now.

I'm working on all changes suggested and will push them at once.

the-real-i9:
I've pushed the changes. But running `testcontainers-go` seems like a cumbersome task on my machine. Sometimes context timeout is exceeded and other times its just left hanging after all tests have passed.

Should I revert back to testing in the GitHub workflow `test-neo4j.yml` file?

neo4j/go.mod
the-real-i9: done

neo4j/config.go
the-real-i9: done

neo4j/config.go
the-real-i9: done

neo4j/neo4j.go
the-real-i9: Leaving it as neo4jstore.

neo4j/neo4j.go
the-real-i9: done

neo4j/neo4j_test.go
the-real-i9: done

neo4j/go.mod
```diff
@@ -0,0 +1,16 @@
+module github.com/gofiber/storage/neo4j/v1
```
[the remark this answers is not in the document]
the-real-i9: fixed

neo4j/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+Used only for optional fields
+
+```go
+var ConfigDefault = Config{
+ TargetBoltURI: "neo4j://localhost",
+ Node:          "fiber_storage",
+ Reset:         false,
+ GCInterval:    10 * time.Second,
+}
```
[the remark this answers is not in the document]
the-real-i9: fixed

gaby:
@the-real-i9 One thing that is missing is Benchmarks. The benchmarks workflow needs to be updated with a step for neo4j

gaby:
The main `README.md` of the repo also needs to reference Neo4j

the-real-i9:
> @the-real-i9 One thing that is missing is Benchmarks. The benchmarks workflow needs to be updated with a step for neo4j

@gaby I'm already using neo4j test container in my test file. The benchmark workflow will run it against other storage packages that use their service container directly in their workflow. Won't this have effect on the benchmark results?

What do you think?

gaby:
> > @the-real-i9 One thing that is missing is Benchmarks. The benchmarks workflow needs to be updated with a step for neo4j
> 
> @gaby I'm already using neo4j test container in my test file. The benchmark workflow will run it against other storage packages that use their service container directly in their workflow. Won't this have effect on the benchmark results?
> 
> What do you think?
> 
> Here's a proof that the benchmark workflow runs the neo4j storage package against other storage packages: Screenshot from 2025-01-31 15-12-00
> 
> And also, the benchmark test workflow gave this error that I don't see what it has to do with me. Screenshot from 2025-01-31 14-46-47

You are correct, the workflow has a default case which hit your benchmarks.

Dont worry about the error, it's related to running in a PR. Once merge we publish the benchmarks to GitHub Pages.

the-real-i9:
@gaby Ok. Thanks.

> The main `README.md` of the repo also needs to reference Neo4j

What about those live test result tag beside them? They reference this main package which isn't the one I forked.

gaby:
> @gaby Ok. Thanks.
> 
> > The main `README.md` of the repo also needs to reference Neo4j
> 
> What about those live test result tag beside them? They reference this main package which isn't the one I forked.

In your repo, add a line here with your driver. https://github.com/gofiber/storage/blob/main/README.md?plain=1#L78

The url should be github.com/gofiber/storage

README.md
```diff
@@ -67,6 +67,7 @@ type Storage interface {
 - [MSSQL](./mssql/README.md) <a href="https://github.com/gofiber/storage/actions?query=workflow%3A%22Tests+MSSQL%22"> <img src="https://img.shields.io/github/actions/workflow/status/gofiber/storage/test-mssql.yml?branch=main&label=%F0%9F%A7%AA%20&style=flat&color=75C46B" /> </a>
 - [MySQL](./mysql/README.md) <a href="https://github.com/gofiber/storage/actions?query=workflow%3A%22Tests+MySQL%22"> <img src="https://img.shields.io/github/actions/workflow/status/gofiber/storage/test-mysql.yml?branch=main&label=%F0%9F%A7%AA%20&style=flat&color=75C46B" /> </a>
 - [NATS](./nats/README.md) <a href="https://github.com/gofiber/storage/actions?query=workflow%3A%22Tests%20Nats%20Driver%22"> <img src="https://img.shields.io/github/actions/workflow/status/gofiber/storage/test-nats.yml?branch=main&label=%F0%9F%A7%AA%20&style=flat&color=75C46B" /> </a>
+- [Neo4j](./neo4j/README.md)
```
gaby, line 70: You are missing the html tag to the workflow tag.

the-real-i9:
@gaby Done.

```md
- [Neo4j](./neo4j/README.md) <a href="https://github.com/gofiber/storage/actions?query=workflow%3A%22Tests+Neo4j%22"> <img src="https://img.shields.io/github/actions/workflow/status/gofiber/storage/test-neo4j.yml?branch=main&label=%F0%9F%A7%AA%20&style=flat&color=75C46B" /> </a>
```

the-real-i9:
@gaby Thanks for this opportunity. And this conversation.

gaby:
> @gaby Thanks for this opportunity. And this conversation.

Will review today, thanks!

neo4j/config.go
```diff
@@ -0,0 +1,102 @@
+package neo4jstore
```
gaby, line 1: @the-real-i9 I was looking at the other drivers and this can be named `neo4j`. For example our redis one is named `package redis` same as the upstream one and it doesn't collide.

neo4j/neo4j.go
```diff
@@ -0,0 +1,223 @@
...
+	var err error
+	db := cfg.DB
+	if db == nil {
+		db, err = newDriverWithContext(neo4jConnConfig{
+			URI:            cfg.URI,
+			Auth:           cfg.Auth,
+			Configurations: cfg.Configurations,
+		})
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Unable to create connection pool: %v\n", err)
```
gaby, line 60: This should throw a panic since we can't create connection. That's the behavior in all the drivers.

This can be easily tested using `testify`

neo4j/neo4j.go
```diff
@@ -0,0 +1,223 @@
...
+	ctx := context.Background()
+
+	if err := db.VerifyConnectivity(ctx); err != nil {
+		panic(err)
+	}
+
+	// truncate node if reset set to true
+	if cfg.Reset {
+		if _, err := neo4j.ExecuteQuery(ctx, db, fmt.Sprintf("MATCH (n:%s) DELETE n FINISH", cfg.Node), nil, neo4j.EagerResultTransformer); err != nil {
+			db.Close(ctx)
```
gaby, line 73: Add a log message here

neo4j/neo4j.go
```diff
@@ -0,0 +1,223 @@
...
+			panic(err)
+		}
+	}
+
+	// create index on key
+	if _, err := neo4j.ExecuteQuery(ctx, db, fmt.Sprintf("CREATE INDEX neo4jstore_key_idx IF NOT EXISTS FOR (n:%s) ON (n.k)", cfg.Node), nil, neo4j.EagerResultTransformer); err != nil {
+		db.Close(ctx)
+		panic(err)
```
gaby, line 81: Add a log message here

neo4j/neo4j.go
```diff
@@ -0,0 +1,223 @@
...
+	// result model
+	var model model
+	mapToStruct(data, &model)
+
+	// If the expiration time has already passed, then return nil
+	if model.Exp != 0 && model.Exp <= time.Now().Unix() {
+		return nil, nil
+	}
+
+	return utils.UnsafeBytes(model.Val), nil
```
gaby, line 132: Why are we using UnsafeBytes here?

neo4j/neo4j.go
```diff
@@ -0,0 +1,223 @@
...
+// gc deletes all expired entries
+func (s *Storage) gc(t time.Time) {
+	_, _ = neo4j.ExecuteQuery(context.Background(), s.db, s.cypherGC, map[string]any{"exp": t.Unix()}, neo4j.EagerResultTransformer)
+}
+
+func mapToStruct(src map[string]any, dest any) {
+	bt, _ := json.Marshal(src)
+
+	if err := json.Unmarshal(bt, dest); err != nil {
+		panic(err)
```
gaby, line 221: This should return an error instead of panic as suggested by @coderabbitai 

If we can't unmarshall the data, it will cause the whole server to crash.

neo4j/neo4j_test.go
```diff
@@ -0,0 +1,219 @@
...
+var testStore *Storage
+
+// TestMain sets up and tears down the test container
+func TestMain(m *testing.M) {
+	ctx := context.Background()
+
+	// Start a Neo4j test container
+	neo4jContainer, err := neo4j.Run(ctx,
+		"neo4j:5.26",
```
gaby, line 22: Move this to a `const` outside this function. That way it's easier to find/update in the future.

gaby (changes requested):
Minor thing, overall looks good.

neo4j/neo4j_test.go
```diff
@@ -0,0 +1,219 @@
...
+)
+
+var testStore *Storage
+
+// TestMain sets up and tears down the test container
+func TestMain(m *testing.M) {
+	ctx := context.Background()
+
+	// Start a Neo4j test container
+	neo4jContainer, err := neo4j.Run(ctx,
```
mdelapenya, line 21: I like the idea of starting with one container per test, so instead of having a TestMain function, having a function that creates the Neo4j store, and calling it por test would be my choice. It will allow parallel execution without the need of taking care of the test data

the-real-i9:
@gaby Now I get it all. It's ok to panic during initializations where the server is yet to start. But never in the code where the server has already started, and users are already sending requests.

I've been trying to get a clear understanding of where to panic and where not to. Thanks!

the-real-i9:
> I like the idea of starting with one container per test, so instead of having a TestMain function, having a function that creates the Neo4j store, and calling it por test would be my choice. It will allow parallel execution without the need of taking care of the test data

@mdelapenya 
Hmm... It takes a sometime for the container to start, you know... I'll do this and compare the time it takes for the whole test to run in both alternatives.

the-real-i9:
> > I like the idea of starting with one container per test, so instead of having a TestMain function, having a function that creates the Neo4j store, and calling it por test would be my choice. It will allow parallel execution without the need of taking care of the test data
> 
> @mdelapenya Hmm... It takes a sometime for the container to start, you know... I'll do this and compare the time it takes for the whole test to run in both alternatives.
> 
> [UPDATE]: Ok. So I tried it. Turns out the simplest test function took around 60s, while others took around 40s. As of the writing the running time is currently at 7mins. on github workflow, and it's not done.
> 
> [UPDATE]: The tests eventually failed due to timeout.

@mdelapenya With a single container in `TestMain`, running all tests took around 20s.

the-real-i9:
@gaby I've addressed all the reviews. Thanks. The PR has been updated.

gaby (approved):
LGTM

ReneWerner87 (approved):

gaby:
@the-real-i9 Thank you 💪

the-real-i9:
> @the-real-i9 Thank you 💪

@gaby @ReneWerner87 🙏 It's an honour to be a part of this.

PR #1515, opened 2024-08-29
chore: adopt testcontainers-go for Postgres, MySQL and MongoDB

mdelapenya:
## What does this PR do?

It uses testcontainers-go in the integration tests for the given stores, refactoring the tests to always consume a new test store, backed by a Docker container.

This PR also bumps testcontainers-go from v0.33.0 to v0.35.0

## Why is it important?
Reproducibility: any developer will be able to run the tests locally exactly equals as in the Github Actions.

## Related issues
For reference, please check out previous work for Couchbase and Minio (#1508) and ClickHouse (#1506)

There is more ongoing work, as defined in #1507

## Other concerns

I started updating MSSQL module, but the container's pid 1 crashes, so I'm going to postpone its addition, using a separate PR for that.

## Summary by CodeRabbit

- **Tests**
  - Enhanced containerized testing across storage modules to ensure each test runs in a fresh, isolated instance.
  - Improved resource cleanup and wait strategies, leading to more robust and reliable test executions.
  - Updated underlying service configurations and images to boost overall stability during testing.
  - Introduced new test functions and constants for better clarity and resource management.
  - Streamlined error handling and resource management in test cases for improved maintainability.

PR #1508, opened 2024-08-21
chore: use testcontainers-go for couchbase and minio

mdelapenya:
- **chore: missing usage of constant**
- **chore: use testcontainers-go module for couchbase in tests**
- **chore: simplify**
- **chore: use testcontainers-go module for minio in tests**

## What does this PR do?
This PR adds testcontainers-go to the Couchbase and Minio modules, updating the current code to be consistent across modules in how the store is created for testing purpose (minio created one container per test suite, instead of per test method).

## Why is it important?
Increase adoption of testcontainers-go for consistency when running the integration tests on contributions.

## Follow-ups
Consider a) running one container per test suite Vs b) one container per test function. While a) is optimal for resource allocation, it can lead to test data polution if tests are not written correctly to create/destroy its data. OTOH b) can lead to have more containers running, but given the type of tests here, I'd say this is meaningless for the current state of the project.

## Summary by CodeRabbit

- **New Features**
	- Introduced environment variables for Couchbase and MinIO to enhance testing infrastructure.
	- Added a new function for setting up a MinIO test storage environment, improving reliability and isolation of tests.

- **Improvements**
	- Streamlined Couchbase and MinIO testing workflows by removing manual setup steps and utilizing pre-defined Docker images.
	- Enhanced error handling in storage setup functions to improve robustness and prevent misleading test results.

- **Refactor**
	- Updated functions for better clarity and maintainability, directly using `testing.TB` type.
	- Simplified test initialization processes across various components.

PR #1506, opened 2024-08-20
feat: use testcontainers-go for clickhouse tests

mdelapenya:
Hello from the [Testcontainers for Go](https://github.com/testcontainers/testcontainers-go/) community :wave:!

This PR is a PoC for adopting testcontainers-go as the testing library for integration testing of the different storage integrations living in this project, for which we do have [Go modules](https://testcontainers.com/modules/?language=go) representing the majority of them. Please let me know if you have any concern about adding it, I'm really happy to assist in addressing them. Of course I'd understand any result for the PR.

## What does this PR do?

This PR adds testcontainers-go to the Clickhouse module, updating the function that expects a local instance running on 127.0.0.1:9001 to be present on the system. Instead, testcontainers-go will start the instance from the test sourcecode.

## Why is it important?
Less setup, as developers will just clone the repository and run the tests (given Docker or another container runtime 100% compatible is installed in the machine). Therefore, the CI experience is exactly the same as running the tests locally, which improves the debuggability.

As a side note, GH actions has Docker already installed, so it should work transparently.

## How to test this PR?

Just go and do like the CI does:

```shell
cd ./clickhouse && go clean -testcache && go test ./... -v -race
```
## Follow-up

If considered positively, add more PRs for the different storages.

## Summary by CodeRabbit

- **Documentation**
	- Enhanced the `README.md` with new sections on testing and local development, including requirements for Docker and usage of Testcontainers for integration tests.
	- Restructured the "Local development" section for improved clarity on setting up a Clickhouse cluster.

- **Tests**
	- Improved the ClickHouse testing suite by integrating dynamic configuration for connection settings using Testcontainers, enhancing test reliability and flexibility.
	- Streamlined the GitHub Actions workflows by removing direct Clickhouse startup steps and introducing environment variables for image reference, simplifying the testing process.

PR #1574, opened 2025-03-19
feat: add testcontainers-go tests to DynamoDB and NATS

mdelapenya:
- **feat: use testcontainers-go's DynamoDB module**
- **feat: use testcontainers-go's nats module**

## What does this PR do?
It follows the same strategy of adding testcontainers-go to make it possible running the integration tests for the storage layer in the local machine,
not needing to install the given storage technology in the host.

I'm copying the result of executing the `gen-test-certs.sh` into the testdata dir of the nats module, although I'm open to discuss whether it's worth it having them statically or dynamically.

## Why is it important?
Run the tests exactly the same in the Ci as in the local machine, making it possible to test/debug locally.

## Related PRs
- #1515
- #1506
- #1508

## Summary by CodeRabbit

- **Tests**
  - Improved test environments with container-based setups for both DynamoDB and NATS, ensuring a more modular and robust testing process.
  - Enhanced error handling during test initialization for more reliable test execution.
  
- **Configuration**
  - Updated TLS settings for NATS by correcting certificate and key file paths to maintain proper security configurations.


nats/nats_test.go
```diff
@@ -1,37 +1,127 @@
...
-		// Enable TLS with client certificate authentication
-		nats.ClientCert("../tls/client.crt", "../tls/client.key"),
-		nats.RootCAs("../tls/ca.crt"),
-	},
-	KeyValueConfig: jetstream.KeyValueConfig{
-		Bucket:  "test",
-		Storage: jetstream.MemoryStorage,
-	},
+const (
+	// mysqlImage is the default image used for running MySQL in tests.
```
mdelapenya, line 22:
```suggestion
	// natsImage is the default image used for running NATS in tests.
```

.github/workflows/benchmark.yml
```diff
@@ -147,9 +135,11 @@ jobs:
           MSSQL_PASSWORD: MsSql!1234
           TEST_CLICKHOUSE_IMAGE: "clickhouse/clickhouse-server:23-alpine"
           TEST_COUCHBASE_IMAGE: "couchbase:enterprise-7.6.5"
+          TEST_DYNAMODB_IMAGE: amazon/dynamodb-local:2.6.0
```
gaby, line 138: Leave this as latest. Maintaining track of patch releases is a pain 😂

dynamodb/dynamodb_test.go
```diff
@@ -1,30 +1,54 @@
...
 
 	"github.com/stretchr/testify/require"
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/modules/dynamodb"
 )
 
-var testStore *Storage
+const (
+	// dynamoDBImage is the default image used for running DynamoDB in tests.
+	dynamoDBImage              = "amazon/dynamodb-local:2.6.0"
```
gaby, line 15: Same here

nats/nats_test.go
```diff
@@ -1,37 +1,127 @@
...
+
+var (
+	//go:embed testdata/nats-tls.conf
+	natsTLSConfig string
+
+	//go:embed testdata/tls/ca.crt
+	caCrt string
+
+	//go:embed testdata/tls/redis.crt
+	redisCrt string
```
gaby, line 36: Why is this named redisCrt?

gaby (changes requested):

.github/workflows/benchmark.yml
mdelapenya: The issue with latest is that it could be anything causing tests to break. OTOH I see the value in not maintaining the versions of all supported technologies. Changing it in a follow-up commit.

nats/nats_test.go
gaby: Yeah, we should rename those files. I think it's because it was reusing the ones from Redis storage. 

There's a script in this repo for generating certs, it might be hardcoded to use "redis" as name 😂

mdelapenya: Would you see it with good eyes to use https://github.com/mdelapenya/tlscert to generate the cert files on the fly? Something like:

```go
func createTLSCerts(t testing.TB) (*tlscert.Certificate, *tlscert.Certificate, *tlscert.Certificate) {
	t.Helper()

	tmpDir := t.TempDir()

	// Generate CA certificate
	caCert := tlscert.SelfSignedFromRequest(tlscert.Request{
		Host:        "localhost",
		IPAddresses: []net.IP{net.ParseIP("127.0.0.1")},
		Name:        "ca",
		IsCA:        true,
		ParentDir:   tmpDir,
	})
	require.NotNil(t, caCert)

	// Generate client certificate
	clientCert := tlscert.SelfSignedFromRequest(tlscert.Request{
		Host:        "localhost",
		IPAddresses: []net.IP{net.ParseIP("127.0.0.1")},
		Name:        "client",
		Parent:      caCert,
		ParentDir:   tmpDir,
	})
	require.NotNil(t, clientCert)

	// Generate nats certificate
	natsCert := tlscert.SelfSignedFromRequest(tlscert.Request{
		Host:        "localhost",
		IPAddresses: []net.IP{net.ParseIP("127.0.0.1")},
		Name:        "nats",
		Parent:      caCert,
		ParentDir:   tmpDir,
	})
	require.NotNil(t, natsCert)

	return caCert, clientCert, natsCert
}
```

the container files:
```go
Files: []testcontainers.ContainerFile{
					{
						Reader:            bytes.NewReader(ca.Bytes),
						ContainerFilePath: "/tls/ca.crt",
						FileMode:          0o0644,
					},
					{
						Reader:            bytes.NewReader(natsCert.Bytes),
						ContainerFilePath: "/tls/nats.crt",
						FileMode:          0o0644,
					},
					{
						Reader:            bytes.NewReader(natsCert.KeyBytes),
						ContainerFilePath: "/tls/nats.key",
						FileMode:          0o0644,
					},
				},
```

the nats client:
```go
return New(Config{
		URLs: uri,
		NatsOptions: []nats.Option{
			nats.MaxReconnects(2),
			// Enable TLS with client certificate authentication
			nats.ClientCert(client.CertPath, client.KeyPath),
			nats.RootCAs(ca.CertPath),
		},
		KeyValueConfig: jetstream.KeyValueConfig{
			Bucket:  "test",
			Storage: jetstream.MemoryStorage,
		},
	}), nil
```

gaby: @mdelapenya Sounds good. That way the shell script can go away 😂

mdelapenya: everything as code!

dynamodb/dynamodb_test.go
```diff
@@ -109,48 +157,64 @@ func Test_DynamoDB_Reset(t *testing.T) {
...
+	require.NoError(b, err)
+	defer testStore.Close()
+
 	for i := 0; i < b.N; i++ {
-		_ = testStore.Set("john", []byte("doe"), 0)
+		err = testStore.Set("john", []byte("doe"), 0)
+		require.NoError(b, err)
+
 		err = testStore.Delete("john")
+		require.NoError(b, err)
```
mdelapenya, line 218: Mmm I think this is incorrect, as it could impact the benchmarks. Changing that to have an slice of errors that will be evaluated outside of the benchmark's for loop

dynamodb/dynamodb_test.go
```diff
@@ -109,48 +157,75 @@ func Test_DynamoDB_Reset(t *testing.T) {
...
+	err = testStore.Set("john", []byte("doe"), 0)
 	require.NoError(b, err)
 
 	b.ReportAllocs()
 	b.ResetTimer()
 
+	errs := make([]error, b.N)
 	for i := 0; i < b.N; i++ {
 		_, err = testStore.Get("john")
+		errs[i] = err
```
gaby, line 206: We don't need this here, since these are benchmarks not tests.

dynamodb/dynamodb_test.go
```diff
@@ -109,48 +157,75 @@ func Test_DynamoDB_Reset(t *testing.T) {
...
-	var err error
+	testStore, err := newTestStore(b)
+	require.NoError(b, err)
+	defer testStore.Close()
+
+	errs := make([]error, b.N)
 	for i := 0; i < b.N; i++ {
 		_ = testStore.Set("john", []byte("doe"), 0)
-		err = testStore.Delete("john")
+		errs[i] = testStore.Delete("john")
```
gaby, line 225: Same here

nats/nats_test.go
```diff
@@ -209,62 +345,76 @@ func Test_Storage_Nats_Reset(t *testing.T) {
...
+	err = testStore.Set("john", []byte("doe"), 0)
 	require.NoError(b, err)
 
 	b.ReportAllocs()
 	b.ResetTimer()
 
+	errs := make([]error, b.N)
 	for i := 0; i < b.N; i++ {
 		_, err = testStore.Get("john")
+		errs[i] = err
```
gaby, line 395: This can remove this too

nats/nats_test.go
```diff
@@ -209,62 +345,76 @@ func Test_Storage_Nats_Reset(t *testing.T) {
...
 
 	b.ReportAllocs()
 	b.ResetTimer()
 
-	var err error
+	errs := make([]error, b.N)
 	for i := 0; i < b.N; i++ {
 		_ = testStore.Set("john", []byte("doe"), 0)
-		err = testStore.Delete("john")
+		errs[i] = testStore.Delete("john")
```
gaby, line 414: Same here

gaby (changes requested):
LGTM, just a few comments

dynamodb/dynamodb_test.go
mdelapenya: question: I'm not that familiar with how benchmarks deal with errors. Would the for loop be hiding errors that happen in between, just capturing if the last error was a real error?

gaby: You have 3 options:
- ignore the errors
- capture the last one and check outside
- add a separate benchmark that checks on each loop (this will be very slow but validates that each loop is working as intended.

I have implemented the last case in the templates repo, it was used to find a data race issue. See here https://github.com/gofiber/template/blob/master/html/html_test.go

For storage the DB itself handles concurrency, we should be fine with just ignoring the error.

If we wanted to in the future all the benchmarks can be updated to look like the templates repo

mdelapenya: Cool, thanks for the explanation, will update this PR now

mdelapenya:
@gaby now that I'm part of the org, what's the policy for merging: waiting for +2 approvals? Is it any of you, me the one merging?

PR #1638, opened 2025-04-08
🔥 Feature: Added SurrealDB 

aliyasirnac:
This PR adds a new storage driver for SurrealDB.

- Implements the Storage interface
- Supports expiration (TTL)
- Includes full tests and benchmark
- Adds README and configuration

Note: Since SurrealDB SDK is generic, a custom `List()` was implemented to fetch all entries.


.github/workflows/test-surrealdb.yml
```diff
@@ -0,0 +1,41 @@
...
+    paths:
+      - 'surrealdb/**'
+name: "Tests SurrealDB"
+jobs:
+  Tests:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        go-version:
+          - 1.22.x
```
gaby, line 18: Remove 1.22.x

.github/workflows/test-surrealdb.yml
```diff
@@ -0,0 +1,41 @@
...
+        command: [ "start", "--user", "root", "--pass", "root" ]
+    steps:
+      - name: Fetch Repository
+        uses: actions/checkout@v4
+      - name: Install Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '${{ matrix.go-version }}'
+
+      - name: Run ScyllaDb
```
gaby, line 35: Fix name

surrealdb/README.md
```diff
@@ -0,0 +1,105 @@
...
+func (s *Storage) Get(key string) ([]byte, error)
+func (s *Storage) Set(key string, val []byte, exp time.Duration) error
+func (s *Storage) Delete(key string) error
+func (s *Storage) Reset() error
+func (s *Storage) Close() error
+func (s *Storage) Conn() *gocb.Cluster
+func (s *Storage) List() ([]byte, error) {
+```
+### Installation
+SurrealDB is tested on Go 1.22 and 1.23 with support for modules.
```
gaby, line 33: Tested on the latest two version of Golang

surrealdb/README.md
```diff
@@ -0,0 +1,105 @@
...
+title: SurrealDB 
+---
+
+![Release](https://img.shields.io/github/v/tag/gofiber/storage?filter=surrealdb*)
+[![Discord](https://img.shields.io/discord/704680098577514527?style=flat&label=%F0%9F%92%AC%20discord&color=00ACD7)](https://gofiber.io/discord)
+![Test](https://img.shields.io/github/actions/workflow/status/gofiber/storage/test-surrealdb.yml?label=Tests)
+
+A SurrealDB storage driver using [surrealdb/surrealdb.go](https://github.com/surrealdb/surrealdb.go).
+
+**Note: Requires Go 1.20 and above**
```
gaby, line 12: Remove this line

surrealdb/README.md
```diff
@@ -0,0 +1,105 @@
...
+func (s *Storage) Reset() error
+func (s *Storage) Close() error
+func (s *Storage) Conn() *gocb.Cluster
+func (s *Storage) List() ([]byte, error) {
+```
+### Installation
+SurrealDB is tested on Go 1.22 and 1.23 with support for modules.
+Make sure to initialize a Go module first if you haven’t already:
+```bash
+go get github.com/gofiber/storage/surrealdb/v2
```
gaby, line 36: Remove /v2 since its the first release

surrealdb/README.md
```diff
@@ -0,0 +1,105 @@
...
+```
+
+### Examples
+Import the storage package.
+```go
+import "github.com/gofiber/storage/surrealdb/v2"
```
gaby, line 42: Same here

surrealdb/go.mod
```diff
@@ -0,0 +1,20 @@
+module github.com/gofiber/storage/surrealdb/v2
+
+go 1.22
+
```
gaby, line 4: Use go 1.23.0

surrealdb/go.mod
```diff
@@ -0,0 +1,20 @@
+module github.com/gofiber/storage/surrealdb/v2
```
gaby, line 1: Remove /v2

surrealdb/surrealdb_test.go
```diff
@@ -0,0 +1,121 @@
+package surrealdb
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/stretchr/testify/require"
+	"os"
+	"testing"
+	"time"
```
gaby, line 9: Tests should use `testcontainers-go`. See some of the other packages

surrealdb/surrealdb_test.go
```diff
@@ -0,0 +1,121 @@
...
+func BenchmarkSet(b *testing.B) {
+	store, err := New(ConfigDefault)
+	if err != nil {
+		b.Fatalf("failed to init storage: %v", err)
+	}
+	defer store.Close()
+
+	for i := 0; i < b.N; i++ {
+		key := fmt.Sprintf("bench-key-%d-%d", i, time.Now().UnixNano())
+		err := store.Set(fmt.Sprintf("bench-key-%s", key), []byte("value"), 0)
```
gaby, line 116: Remove check for err, its a benchmark

surrealdb/surrealdb_test.go
```diff
@@ -0,0 +1,121 @@
...
+		if err != nil {
+			b.Errorf("Set failed: %v", err)
+		}
+	}
+}
```
gaby, line 121: Add benchmark for Get

surrealdb/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Signatures
+```go
+func New(config ...Config) (*Storage, error) 
+func (s *Storage) Get(key string) ([]byte, error)
+func (s *Storage) Set(key string, val []byte, exp time.Duration) error
+func (s *Storage) Delete(key string) error
+func (s *Storage) Reset() error
+func (s *Storage) Close() error
+func (s *Storage) Conn() *gocb.Cluster
```
gaby, line 29: Is that the name of the client class? gocb.Clustdr?

gaby: This should be `surrealdb.DB`

surrealdb/surrealdb.go
```diff
@@ -0,0 +1,136 @@
...
+func New(config ...Config) (*Storage, error) {
+	cfg := configDefault(config...)
+	db, err := surrealdb.New(cfg.ConnectionString)
+	if err != nil {
+		return nil, err
+	}
+
+	if err = db.Use(cfg.Namespace, cfg.Database); err != nil {
+		return nil, err
+	}
```
gaby, line 41: If we cant init, configure or auth, we should panic

gaby (changes requested):
Looks good overall, just a few comments here and there. Will re-review once those are done. Thanks! 💪

aliyasirnac:
@gaby thanks for your feedbacks

.github/workflows/test-surrealdb.yml
```diff
@@ -0,0 +1,40 @@
...
+  Tests:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        go-version:
+          - 1.23.x
+          - 1.24.x
+    services:
+      surrealdb:
+        image: 'surrealdb/surrealdb:latest'
```
gaby, line 22: You are running surrealdb twice in this workflow.

surrealdb/README.md
```diff
@@ -0,0 +1,101 @@
+---
+id: surrealdb 
+title: SurrealDB
```
gaby, line 3: remove trailing spaces from line 2 and 3

surrealdb/go.mod
```diff
@@ -0,0 +1,68 @@
+module github.com/gofiber/storage/surrealdb
+
+go 1.23.0
+
+toolchain go1.23.3
```
gaby, line 5: remove toolchain line

surrealdb/surrealdb_test.go
```diff
@@ -0,0 +1,205 @@
...
+	require.NotContains(t, result, "expired")
+}
+
+func BenchmarkSet(b *testing.B) {
+	testStore, err := newTestStore(b)
+	require.NoError(b, err)
+	defer testStore.Close()
+
+	for i := 0; i < b.N; i++ {
+		key := fmt.Sprintf("bench-key-%d-%d", i, time.Now().UnixNano())
```
gaby, line 182: Can you pick a more simple or static key? This will perform very bad compared to the other drivers.

gaby:
@aliyasirnac SurrealDB allows running in KV and InMemoryMode. I think in those modes you don't have to login?

Can we add logic to take those into account, we probably need separate benchmarks for those too.

See here: https://github.com/surrealdb/surrealdb.go?tab=readme-ov-file#using-surrealkv-and-memory

.github/workflows/benchmark.yml
```diff
@@ -109,7 +109,11 @@ jobs:
         run: |
           docker run --name scylladb -p 9042:9042 -p 19042:19042 -p 9160:9160 -p 7000:7000 -p 7001:7001 -p 7199:7199 -p 9180:9180 -d scylladb/scylla:latest --broadcast-address 127.0.0.1 --listen-address 0.0.0.0 --broadcast-rpc-address 127.0.0.1
           sleep 15 # Wait for ScyllaDb to initialize
-
+      - name: Install SurrealDB
```
mdelapenya, line 112: There exists a surrealdb module in testcontainers-go. I think we must use it here

surrealdb/surrealdb_test.go
mdelapenya: Reference: https://testcontainers.com/modules/surrealdb/?language=go

aliyasirnac:
> @aliyasirnac SurrealDB allows running in KV and InMemoryMode. I think in those modes you don't have to login?
> 
> Can we add logic to take those into account, we probably need separate benchmarks for those too.
> 
> See here: https://github.com/surrealdb/surrealdb.go?tab=readme-ov-file#using-surrealkv-and-memory

I’m currently implementing your suggested approach, but I’m still debating how to structure the test file. I have two scenarios in mind:

1. **Single `newTestStore` Function**  
   I could have one function, `newTestStore(t testing.TB, mode string)`, which uses a switch statement on `config.Mode`. For example, if `mode` is `"kv"` or `"memory"`, it would skip the login; if `mode` is `"default"`, it would perform a login. Then I would write my tests like this:

   ```go
   func Test_Surrealdb_Modes(t *testing.T) {
       modes := []string{"default", "memory", "kv"}

       for _, mode := range modes {
           t.Run(mode, func(t *testing.T) {
               t.Run("Create", func(t *testing.T) {
                   // test case...
               })
           })
       }
   }
   
2. Separate Functions per Mode
Alternatively, I could write a dedicated function for each mode, such as newTestStoreMemory, newTestStoreKV, etc. Then I’d create separate test functions for each store/mode.

Which approach do you think is better?

MitulShah1:
Since I was working on same but PR already raised , I have one suggestion to follow as per most driver in fiber

1. As we have already predefined scenarios, and its key,value pair and model is already pre defined in driver, can you add Initialise schema and check schema while initialise surrealDB?

like

`
var (
	initQuery = `DEFINE TABLE %s SCHEMAFULL;
		 DEFINE FIELD k ON %s TYPE string;
		 DEFINE FIELD v ON %s TYPE string;
		 DEFINE FIELD e ON %s TYPE int;`
	checkSchemaQuery = `INFO FOR TABLE %s;`
)

func New(config ...Config) *Storage {

//Your code...

       // Initialize the storage
	query := fmt.Sprintf(initQuery, cfg.DefaultTable, cfg.DefaultTable, cfg.DefaultTable, cfg.DefaultTable)
	if err := s.db.Send(&res, "query", query, nil); err != nil {
		panic(err)
	}

	// Check schema
	if err := s.db.Send(&res, "query", fmt.Sprintf(checkSchemaQuery, cfg.DefaultTable), nil); err != nil {
		panic(err)
	}
`

2. Can we add Garbage collection configuration?

like

`
type Config struct {
....

// Time before deleting expired keys
//
GCInterval time.Duration
}

func New(config ...Config) *Storage {

.....

// Start garbage collector
go store.gcTicker()
}

// Garbage collector to delete expired keys
func (s *Storage) gcTicker() {
	ticker := time.NewTicker(s.config.GCInterval)
	defer ticker.Stop()

	for range ticker.C {
		var res QueryResponse
		params := map[string]interface{}{
			"expiry": time.Now().Unix(),
		}

		if err := s.db.Send(&res, "query", "DELETE FROM %s WHERE e <= $expiry AND e != 0", params); err != nil {
			// Just log the error and continue
			fmt.Printf("SurrealDB GC error: %v\n", err)
		}
	}
}
`

3. Reset is also missing.

**Note**: Above is my suggestions to keep code same as other driver follow in storage. rest is depend on PR reviewer :)

gaby:
@MitulShah1 Next time, comment on the issue as it was already assigned to @aliyasirnac Thanks 💪

MitulShah1:
Thanks @gaby, i take care next rime, i just suggested changes as per repo 😀 its always upto reviewer no force 😀

aliyasirnac:
Hi @gaby,
You previously suggested updating the code to support SurrealDB’s KV and in-memory modes. However, the current surrealdb.go SDK does not support these modes at the moment.

According to [this issue](https://github.com/surrealdb/surrealdb.go/issues/197), embedded database support (including memory:// and surrealkv://) is currently not implemented in the Go SDK. They’re waiting for the C SDK to be developed in order to bring this functionality.

You can also see in [this part of the code](https://github.com/surrealdb/surrealdb.go/blob/8b9d3abc6428432698fea81ed28691127cd50057/db.go#L53-L54) that only HTTP and WebSocket protocols are supported right now.

So unfortunately, I’m currently unable to proceed with the update as requested until embedded support is added.

gaby:
@aliyasirnac Should we add WebSocket support then? Usually that gives higher throughput than HTTP. If its minimal changes, else lets stick with HTTP

aliyasirnac:
> @aliyasirnac Should we add WebSocket support then? Usually that gives higher throughput than HTTP. If its minimal changes, else lets stick with HTTP

Yes, currently when starting the test container, we build the connection string using something like fmt.Sprintf("ws://%s:%s", host, port), and everything works as expected.

I tried switching to an HTTP connection instead (i.e. http://), but it results in an error — likely because the test container setup only supports WebSocket connections at the moment. So for now, WebSocket appears to be the only working option in this test environment.

mdelapenya:
@aliyasirnac if you see any limitation in the testcontainers implementation, please raise an issue in https://github.com/testcontainers/testcontainers-go/issues 🙏

aliyasirnac:
> @aliyasirnac if you see any limitation in the testcontainers implementation, please raise an issue in https://github.com/testcontainers/testcontainers-go/issues 🙏

I think I analyzed it wrong because I just did a few tests and when I log in with `surrealdb.New(“http://localhost:8000”)` it connects to the database but when I do a data creation it gives the following error: `panic: cbor: 18 bytes of extraneous data starting at index 21` and when I try the same code with `ws://localhost:8000`, the data insertion is successful. I don't think this is a problem with testcontainers.

aliyasirnac:
> > @aliyasirnac if you see any limitation in the testcontainers implementation, please raise an issue in https://github.com/testcontainers/testcontainers-go/issues 🙏
> 
> I think I analyzed it wrong because I just did a few tests and when I log in with `surrealdb.New(“http://localhost:8000”)` it connects to the database but when I do a data creation it gives the following error: `panic: cbor: 18 bytes of extraneous data starting at index 21` and when I try the same code with `ws://localhost:8000`, the data insertion is successful. I don't think this is a problem with testcontainers.

I think we need to tell the user to use websocket.

gaby:
> > > @aliyasirnac if you see any limitation in the testcontainers implementation, please raise an issue in https://github.com/testcontainers/testcontainers-go/issues 🙏
> > 
> > 
> > I think I analyzed it wrong because I just did a few tests and when I log in with `surrealdb.New(“http://localhost:8000”)` it connects to the database but when I do a data creation it gives the following error: `panic: cbor: 18 bytes of extraneous data starting at index 21` and when I try the same code with `ws://localhost:8000`, the data insertion is successful. I don't think this is a problem with testcontainers.
> 
> I think we need to tell the user to use websocket.

My guess is this is a bug with testcontainers? @mdelapenya

surrealdb/README.md
```diff
@@ -0,0 +1,101 @@
+---
+id: surrealdb
```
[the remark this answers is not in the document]
ReneWerner87: @aliyasirnac don't forget this

.github/workflows/benchmark.yml
```diff
@@ -154,4 +154,4 @@ jobs:
           auto-push: ${{ github.event_name == 'push' || github.event_name == 'workflow_dispatch' }}
           save-data-file: ${{ github.event_name == 'push' || github.event_name == 'workflow_dispatch' }}
 
-      
\ No newline at end of file
+
```
mdelapenya, line 157: bug: we are missing the `TEST_SURREALDB_IMAGE` declaration in the run benchmarks job. I know the code use defaults, but I'd advocate for readability, so this file is self-contained. Please add it here: https://github.com/gofiber/storage/pull/1638/files#diff-aacbd0338bcf598920c120b69adaba2672acc7ea314cfca0f792bebe5fe2acd0R135

.github/workflows/test-surrealdb.yml
```diff
@@ -0,0 +1,29 @@
...
+    steps:
+      - name: Fetch Repository
+        uses: actions/checkout@v4
+      - name: Install Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '${{ matrix.go-version }}'
+
+      - name: Run Test
+        run: cd ./surrealdb && go test ./... -v -race
```
mdelapenya, line 29: bug: same as above
```suggestion
        env:
          TEST_SURREALDB_IMAGE: surrealdb/surrealdb:latest  
        run: cd ./surrealdb && go test ./... -v -race
```

surrealdb/surrealdb_test.go
```diff
@@ -0,0 +1,220 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	port, err := surrealdbContainer.MappedPort(context.Background(), "8000")
+	if err != nil {
+		return nil, err
+	}
+
+	url := fmt.Sprintf("ws://%s:%s", host, port.Port())
```
mdelapenya, line 47: suggestion: simplify using module's API
```suggestion
	url, err := surrealdbContainer.URL(ctx)
	if err != nil {
		return nil, err
	}
```

surrealdb/surrealdb_test.go
```diff
@@ -0,0 +1,220 @@
...
+)
+
+func newTestStore(t testing.TB) (*Storage, error) {
+	t.Helper()
+
+	img := surrealDb
+	if imgFromEnv := os.Getenv(surrealDbImageEnvVar); imgFromEnv != "" {
+		img = imgFromEnv
+	}
+	surrealdbContainer, err := surrealdb.Run(context.Background(), img)
```
mdelapenya, line 30: suggestion: pass credentials to the container using the existing options
```suggestion
	surrealdbContainer, err := surrealdb.Run(
		context.Background(), img,
		surrealdb.WithUsername(surrealDbUser),
		surrealdb.WithPassword(surrealDbPass),
	)
```

surrealdb/surrealdb_test.go
```diff
@@ -0,0 +1,220 @@
...
+	require.NoError(t, err)
+	defer testStore.Close()
+
+	err = testStore.Set("test", []byte("test12345"), 0)
+	require.NoError(t, err)
+
+	get, err := testStore.Get("test")
+	require.NoError(t, err)
+	require.NotEmpty(t, get)
+
```
mdelapenya, line 81: nit
```suggestion
```

surrealdb/surrealdb_test.go
```diff
@@ -0,0 +1,220 @@
...
+func Test_Surrealdb_ListSkipsExpired(t *testing.T) {
+	testStore, err := newTestStore(t)
+	require.NoError(t, err)
+	defer testStore.Close()
+
+	_ = testStore.Set("valid", []byte("123"), 0)
+
+	_ = testStore.Set("expired", []byte("456"), 1*time.Second)
+	time.Sleep(2 * time.Second)
```
mdelapenya, line 160: question: why 2 seconds and no other time? Will this value change depending on the machine the tests are run, like devs laptops or the GH workers?

mdelapenya (commented):
Added some comments regarding the usage of testcontainers-go

Thanks @aliyasirnac for your contribution 🙇

surrealdb/surrealdb_test.go
aliyasirnac: The reason I added time.sleep here was to guarantee that the value we set will expire. I don't understand if this value should change according to the machine and not 2 seconds?

gaby: This is fine, since ttl is 1s

gaby:
@aliyasirnac What's left on this PR?

aliyasirnac:
> @aliyasirnac What's left on this PR?

I don't have much of an idea. KV and in-memory are not currently supported in the official sdk, so I couldn't add them. Some repositories had garbage collectors, maybe I can add them and maybe we can add and refactor the bots' suggestions.

surrealdb/go.mod
```diff
@@ -0,0 +1,67 @@
+module github.com/gofiber/storage/surrealdb
+
+go 1.23.0
+
+
+require (
+	github.com/stretchr/testify v1.10.0
+	github.com/surrealdb/surrealdb.go v0.3.2
```
mdelapenya, line 8: bug: please run `go mod tidy` in order to sync the file. Also, could you please use testcontainers-go `v0.37.0`? It was released the last week 🙏

surrealdb/surrealdb_test.go
```diff
@@ -0,0 +1,253 @@
...
+
+var (
+	// surrealDb is the default image used for running surrealdb in tests.
+	surrealDb                   = "surrealdb/surrealdb:latest"
+	surrealDbImageEnvVar string = "TEST_SURREALDB_IMAGE"
+	surrealDbUser        string = "root"
+	surrealDbPass        string = "root"
+)
+
+func newTestStore(t testing.TB) (*Storage, error) {
```
mdelapenya, line 22: @aliyasirnac may I take over the testcontainers code and add changes on top of yours? I don't want you to feel I'm too picky and convert this in infinite ping-pong 🏓  causing your frustration 🙏

mdelapenya: All storage modules are defining this `newTestStore` function just returning a pointer to `Storage`, using require.Error as part of the internals of the helper function. Could you apply that, updating all references? Each test function will call it, and will defer the closing of the returned store in the test function, so it's clear for eventual readers that each test function closes the storage. There is one single exception, the Test function for closing a storage, which does not defer it but simply checks it can be closed without errors. Please take a look at the recently added redis or arangodb tests 🙏 

Again, thanks for your work here

surrealdb/surrealdb.go
```diff
@@ -0,0 +1,176 @@
...
+func New(config ...Config) *Storage {
+	cfg := configDefault(config...)
+	db, err := surrealdb.New(cfg.ConnectionString)
+	if err != nil {
+		panic(err)
+	}
+
+	if err = db.Use(cfg.Namespace, cfg.Database); err != nil {
+		panic(err)
+	}
```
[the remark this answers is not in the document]
gaby: We need to panic to make user aware DB is down

surrealdb/surrealdb_test.go
gaby: @aliyasirnac I think this is the only thing missing, everything else looks good to me 💪

ReneWerner87 (approved):

gaby (approved):
👍 LGTM

PR #1639, opened 2025-04-09
feat(azurite): use testcontainers-go for testing azurite

mdelapenya:
## What does this PR do?
It follows the same strategy of adding testcontainers-go to make it possible running the integration tests for the Azurite storage layer in the local machine, not needing to install the given storage technology in the host.

## Why is it important?
Run the tests exactly the same in the Ci as in the local machine, making it possible to test/debug locally.

## Related PRs
- #1515
- #1506
- #1508

## Summary by CodeRabbit

- **Tests**
	- Improved test isolation by establishing independent storage instances for each test.
	- Enhanced dynamic configuration of storage endpoints and credentials for a more realistic and robust testing environment.
	- Strengthened error handling during test initialization to ensure higher test reliability and overall quality.
	- Updated test function naming for clarity and consistency.


gaby (commented):
My only comment is that we were using latest image. Having a tagged image is a maintenance burden.

Other than that, the latest is 3.34.0

.github/workflows/test-azureblob.yml
```diff
@@ -15,18 +15,17 @@ jobs:
         strategy:
             matrix:
                 go-version:
-                    - 1.19.x
-                    - 1.20.x
-                    - 1.21.x
+                    - 1.22.x
```
gaby, line 18: May as well remove 1.22, it's End Of Life

mdelapenya: Should I remove it elsewhere testcontainers is used?

gaby: Yes, only 1.23 and 1.24 are getting updates from the Go team

mdelapenya:
> My only comment is that we were using latest image. Having a tagged image is a maintenance burden.

My concern about latest images is that you do not control what goes inside, so pinning a version is a good idea in terms of reproducibility. OTOH we'll be notified in the case the tests are broken, so fine for me to use latest.

.github/workflows/test-azureblob.yml
mdelapenya: In a follow-up PR, should we update the benchmarks version too?

gaby: I forgot why we did 1.21, it was prob latest at that point, or its the one listed in the main go.mod file

gaby (approved):
👍 LGTM

mdelapenya:
Fine for me, thanks!! :rocket:

PR #1662, opened 2025-04-11
chore(benchmarks): use each module's Go version

mdelapenya:
## What does this PR do?
It updates the GH action for the benchmarks to use Go 1.23

## Why is it important?
Use up-to-date Go versions

## Follows Up
Consider bumping the Go version in the root dir' go.mod file and use it in this action


gaby:
I dont think makes sense to change this, since you want the same baseline for all the benchmarks (the go version) ?

mdelapenya:
All right, I then understood it wrong, I though we wanted to use each module's Go version instead, my bad. Will change to the lowest supported version, 1.23 atm

gaby:
Yeah, 1.23 as baseline works. Since its the lowest active supported Go version

gaby:
> > Yeah, 1.23 as baseline works. Since its the lowest active supported Go version
> 
> @gaby I updated it to 1.23. Probably the comment in the file means using root dir's go.mod file? 🤔 It's on 1.19, and we should then bump it to 1.23. Thoughts?

We will do that change for the next big release.

gaby (approved):
👍 LGTM

Issue #967, opened 2023-08-31
🤗 [Question]: Can we add an option to database storages accept a already initialized connection?


mirusky:
### Question Description

For example my app already opens a connection with redis and I'd like to reuse the same connection, since it's simple to create a health check endpoint that's only checks one redis connection instead of 2 or more. 

Currently I'm opening 3 connections since my code opens 1 and I call `redis.New` twice, so I need to get the remaining connections using `store.Conn` and pass it as dependency to my health check controller to check all connections.

I'd like to know if it's possible to add an option to database driven storages like "Conn" that we can pass a already established connection. Something simillar to [GORM/Existing database connection](https://gorm.io/docs/connecting_to_the_database.html#Existing-database-connection)

### Code Snippet (optional)

```go
package main

import (
	"github.com/redis/go-redis/v9"

	fiberredis "github.com/gofiber/storage/redis"
)

func main() {
	url := "redis://localhost:6379?password=hello&protocol=3"
	opts, err := redis.ParseURL(url)
	if err != nil {
		panic(err)
	}
	rdb := redis.NewClient(opts) // Already connected

	store1 := fiberredis.New(redis.Config{URL: config.RedisURL + "/1"}) // New Conn
	store2 := fiberredis.New(redis.Config{URL: config.RedisURL + "/2"}) // New Conn

	// I'm thinking something like this:
        // Ref: https://github.com/redis/go-redis/issues/1018#issuecomment-517928987
	// db1 := rdb.Conn()
	// db1.Select(context.Background(), 1)
	// db2 := rdb.Conn()
	// db2.Select(context.Background(), 2)
	// store1 := fiberredis.New(redis.Config{Conn: db1})
	// store2 := fiberredis.New(redis.Config{Conn: db2})
}
```

### Checklist:

- [X] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [X] I have checked for existing issues that describe my questions prior to opening this one.
- [X] I understand that improperly formatted questions may be closed without explanation.

mirusky:
I've checked and some stores already have this, just a few don't allow it. 

[storage/postgres](https://github.com/gofiber/storage/tree/main/postgres#config) accepts a conn
[storage/mysql](https://github.com/gofiber/storage/tree/main/mysql#config) accepts too

gaby:
[...]

PR #1680, opened 2025-04-18
Add NewFromConnection method for Redis

sky93:
Introduced NewFromConnection to create Storage using an existing Redis client, enhancing flexibility. Added benchmarks to test Redis operations with this method and updated README.

## Summary by CodeRabbit

- **Documentation**
  - Updated Redis storage driver documentation to include creating storage instances from existing Redis connections with usage examples.

- **New Features**
  - Added support for initializing Redis storage directly from an existing Redis client connection.

- **Tests**
  - Added tests verifying Redis storage operations using existing connections.


sixcolors:
I would recommend that if we accept this P.R. for redid we implement it as a standard for other adapters.

grivera64:
> I would recommend that if we accept this P.R. for redis we implement it as a standard for other adapters.

I agree, it would be nice to have this work with any `db` connection.

Since most `db` connections have a variety of differing methods, the only common method that all storage packages expect to use is `storage.Close()`.

If we do decide to implement it as a standard, I think it would be good enough to allow passing in any struct that implements `Close()` as an argument through making some interface called `storage.Conn` or something along those lines.

gaby:
This is related to #967 questions is, what do we do with all the other config options on each Storage driver?

But yes, this needs to be added to all the drivers.

sky93:
I appreciate you taking a look at my proposed changes!

While I understand there may be discussion around standardizing this approach across all databases, I believe accepting this pull request now would still provide clear benefits:

**Flexibility for Redis Users**
Allowing initialization from existing connections addresses a real problem—many developers already have configured Redis clients prepared, and reusing them reduces unnecessary coding. It's a straightforward enhancement that leaves current code and interfaces untouched.

**A Potential Model for Other Drivers**
As noted, establishing consistent handling of pre-made connections throughout various adapters could occur step-by-step. Accepting this would generate a tested example for Redis serving as a model for databases like MySQL, Postgres, and beyond without postponing relief for Redis developers.

**Backwards-Compatible and Low-Risk**
This addition doesn’t break existing usage: developers can still rely on the standard config-based constructor if they wish. The new method simply provides an alternate path for those who already have a Redis client instance.

Looking forward to any final thoughts or suggestions on next steps :)

DanialV:
This feature is essential, since most applications already use Redis connections for various purposes, eliminating the need to create a new one for fiber storage.

redis/redis_test.go
```diff
@@ -539,3 +539,40 @@ func Benchmark_Redis_SetAndDelete(b *testing.B) {
 
 	require.NoError(b, err)
 }
+
+func Benchmark_Redis_WithConnection_SetAndDelete(b *testing.B) {
```
ReneWerner87, line 543: @sky93 Can you also add a test method with the new possibility

redis/README.md
```diff
@@ -21,6 +21,7 @@ A Redis storage driver using [go-redis/redis](https://github.com/go-redis/redis)
 ### Signatures
 ```go
 func New(config ...Config) Storage
+func NewFromConnection(conn redis.UniversalClient) Storage
```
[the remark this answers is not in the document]
sky93: I added an example. Does it look good?

redis/redis_test.go
sky93: I also added another test.

sky93 (commented):
I added requested changes. Hope it helps make the code better.

gaby (approved):
👍 LGTM

ReneWerner87 (approved):

PR #1723, opened 2025-04-30
feat(redis): use testcontainers in rueidis and valkey

mdelapenya:
## What does this PR do?
It creates a `testhelpers/redis` test module, abstracting the current test implementation for running Redis with testcontainers. This test module has been applied to redis, rueidis and valkey, as they all start a redis/valkey container for testing.

## Why is it important?
Do not repeat the same code for starting redis/valkey across three different modules.

## Other concerns
Please let me know if this is too over-engineered 🙏 so I can step back and provide a simpler solution. I chose this way because there are multiple combinations for creating redis in the current tests: with/out TLS, using secure/unsecure schema, using/not using MTLS, using URL, using HostPort...

## Summary by CodeRabbit
- **New Features**
  - Introduced a reusable helper for running Redis containers in tests, supporting flexible configuration (TLS, connection modes, image selection).

- **Refactor**
  - Updated Redis, Rueidis, and Valkey test suites to use containerized Redis instances for improved test isolation and cleanup.
  - Replaced static and local Redis configurations with dynamic, per-test containers.
  - Expanded and parameterized TLS-related tests for broader coverage.

- **Chores**
  - Removed obsolete test certificate generation script.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new helper package for running Redis containers in tests, supporting flexible configuration including TLS, image selection, and connection modes.
- **Refactor**
  - Updated Redis-based test suites to use isolated, containerized Redis instances for improved test isolation and resource management.
  - Replaced static and manual Redis test setups with reusable container-based helpers.
  - Expanded TLS and connection option test coverage using parameterized subtests.
  - Improved benchmark tests by using a singleton reusable containerized Redis store.
- **Chores**
  - Removed an obsolete script for generating test TLS certificates.


gaby:
@mdelapenya Only question I had is, does the `testhelpers` have to be a module? Other than that, looks fine to me. Less boilerplate.

mdelapenya:
@gaby I started with a module to avoid pushing the dependencies to the root package. Probably by my lack of knowledge about who is importing this root module. If it's fine to push them to the root, then I'm more than happy to convert it into a package, maybe internal: `./internal/testhelpers/redis`?, and update the references in the redis and friends modules.

ReneWerner87:
@gaby @mdelapenya 
the root go.mod is for the storage interface and all files in the root directory
theoretically you could also use the other sub folders via this package, but nobody should actually do that
so basically for the general parts of the storage repos

mdelapenya:
I refreshed this PR. If you find the helper go module as a non good-practice here, I'm fine in moving to any other place, like an `internal/helpers`. I wanted to have a common structure to start redis-like containers, like redis and valkey, avoiding pushing all the dependencies in the root module.

mdelapenya:
@ReneWerner87 @gaby I did a deep research on why the **valkey** module suffered a huge regression after the changes in this PR (before the singleton container applied to the valkey benchmarks, see 331a84b). Let me explain why I chose to initialise the container just once for the valkey benchmarks and let's validate if this is a valid concern. Thanks Cursor for the support 🤖 :

### Root Cause Analysis

The performance regression in the benchmarks was caused by the interaction between Go's benchmark framework and testcontainers-go's container lifecycle management. Here's what was happening:

**1. Benchmark Framework Behavior:**
- I noticed that, when the store was initialised inside a benchmark function, Go's benchmark framework runs each that function multiple times. I added a fmt.Print to verify that it was call multiple times 🤯 
- On the other hand, with the singleton pattern, that print happened just once. 🤯 
- The framework performs multiple iterations to establish statistical significance
- This means the setup code (before b.ResetTimer()) can be executed multiple times ⬅️ 

**2. Original Implementation Issues:**
- I started with each benchmark creating a new container via newTestStore(b)
- The container was created before b.ReportAllocs() and b.ResetTimer()
- However, the benchmark framework was still running this setup code multiple times.
- This resulted in multiple container starts/cleanups per benchmark. E.g., `Benchmark_Valkey_SetAndDelete` was run 5 times.
- Each container start/cleanup cycle caused:
  - New container creation
  - New connection establishment
  - Resource allocation and cleanup

**3. Impact on Performance:**
- The Valkey Module showed severe impact:
  - Set operation: 387 allocs/op (vs 1 in main branch)
  - SetAndDelete operation: 1590 allocs/op (vs 1 in main branch)
  - High memory usage: 128,374 B/op for Set (vs 39 B/op in main)
- This could be due to the overhead of repeated container lifecycle operations

### Interesting discovery

So in the Redis module:
- A new container is created for each benchmark
- The container is cleaned up after each benchmark
- There is no container reuse
- Yet it maintains good performance

This makes the situation even more interesting:
- Both modules create new containers for each benchmark. Actually, any other storage in the repo.
- Both modules, redis and valkey, use the same testcontainers-go code
- Both modules have identical benchmark structure
- But only the Valkey module shows performance degradation
- In the Main Branch, with a Direct Docker run for valkey, and redis:
  - Valkey client performs excellently
    - Set: 1 alloc/op, 39 B/op
    - Get: 11 allocs/op, 609 B/op
    - SetAndDelete: 1 alloc/op, 108 B/op
- In the `fix-ruedis-benchmarks` Branch, this PR:
  - Redis module: performs well (same as main)
  - Valkey module: shows performance regression
  - But when we add container reuse, Valkey performs well again

### Solution
The solution implemented a singleton pattern with container reuse:

**1. Container Reuse:**
- Uses testcontainers-go's reuse feature
- Container is started once and reused across benchmarks
- Avoids repeated container lifecycle operations

**2. Thread-Safe Singleton:**
- Ensures container is initialized only once
- Thread-safe initialization across benchmark goroutines
- Maintains a single connection throughout the benchmark suite

**3. Removed Cleanup Overhead:**
- Removed defer testStore.Close() from benchmarks
- Container cleanup happens at the end of the test session
- Eliminates repeated connection teardown

### Results
The changes restored performance to optimal levels:
- Set: 1 alloc/op, 39 B/op (vs 387 allocs/op, 128,374 B/op before)
- Get: 11 allocs/op, 583 B/op (slightly better than before)
- SetAndDelete: 1 alloc/op, 70 B/op (vs 1590 allocs/op, 528,947 B/op before)

This solution aligns with the main branch's approach of using a singleton store while leveraging testcontainers-go's container reuse capabilities for efficient testing.

### Key Takeaways:
- Benchmark setup code can run multiple times
- Container lifecycle operations are expensive
- Container reuse can be crucial for benchmark performance
- Thread-safe singleton pattern ensures proper resource management
- The performance regression was not in the client implementation but in the test infrastructure

.github/workflows/benchmark.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       - name: Generate filters
         id: filter-setup
         run: |
-          filters=$(find . -maxdepth 1 -type d ! -path ./.git ! -path . -exec basename {} \; | grep -v '^\.' | awk '{printf "%s: \"%s/**\"\n", $1, $1}')
+          filters=$(find . -maxdepth 1 -type d ! -path ./.git ! -path . ! -path ./testhelpers -exec basename {} \; | grep -v '^\.' | awk '{printf "%s: \"%s/**\"\n", $1, $1}')
```
mdelapenya, line 31: question: the helper tests won't be executed because of this. Have you considered moving to `go.work` so that it's easier to identify the module paths?

ReneWerner87 (approved):

PR #1813, opened 2025-07-06
get presigned url function for minio

little-tonii:
## Summary by CodeRabbit

* **New Features**
  * Added the ability to generate presigned URLs for objects, allowing users to share temporary access links with customizable expiration and download options.

* **Documentation**
  * Updated documentation to include details about the new presigned URL generation feature.

* **Tests**
  * Introduced new tests to ensure correct behavior and error handling for presigned URL generation.

### DIFF
--- a/minio/README.md
+++ b/minio/README.md
@@ -32,6 +32,7 @@ func (s *Storage) CheckBucket() error
 func (s *Storage) CreateBucket() error
 func (s *Storage) RemoveBucket() error
 func (s *Storage) Conn() *minio.Client
+func (s *Storage) GetPresignedURL(key string, expiration time.Duration, isDownload bool) (string, error)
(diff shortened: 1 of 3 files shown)


ReneWerner87:
not sure that this specific functionality fits the storage

little-tonii:
> not sure that this specific functionality fits the storage

im not sure with it too. because i realize that we can get the minio connection from storage and then use the official presigned url function from minio 😂

ReneWerner87:
that was exactly my thought
if the function brought a lot of added value, i would include it
but i don't see that right now

little-tonii:
> that was exactly my thought if the function brought a lot of added value, i would include it but i don't see that right now

im agree with you :>

PR #1927, opened 2025-10-13
🔥 feat: Support for DisableStartupCheck for several drivers

### LINKED ISSUE #1924: 🐛 [Bug]: Redis storage constructor panic on network error
### Bug Description

Redis storage constructor panic on network or flushing errors.
Using `panic` in a public library (especially one intended for integration like `fiber/storage/redis`) breaks Go’s normal error-handling semantics and prevents callers from gracefully handling startup failures.

**Why this is problematic**:
- **Panics in public libraries** break normal error handling and make it impossible for applications to gracefully recover or report failures.
Go libraries should return errors, not crash the process.
- **Implicit IO (PING)** during object creation violates constructor expectations- `New()` should only allocate and configure objects, not perform external network calls.
This makes testing, offline environments, and lazy initialization harder.

### How to Reproduce

1. Call `New()` with address that doesn't exist or not reachable.
2. The process will crash.

### Expected Behavior

The constructor should:
- Return (*Storage, error) instead of panicking.
- Avoid performing an implicit PING during initialization.
- Let the caller decide whether to verify connectivity or defer it.

### Storage package Version

v3.4.1

### Code Snippet (optional)

```go
// New creates a new Redis storage instance.
func New(config ...Config) *Storage {
	// Set default config

	// ........

	// Test connection
	if err := db.Ping(context.Background()).Err(); err != nil {
		panic(err)
	}

	// Empty collection if Clear is true
	if cfg.Reset {
		if err := db.FlushDB(context.Background()).Err(); err != nil {
			panic(err)
		}
	}

	// Create new store
	return &Storage{
		db: db,
	}
}
```

### Checklist:

- [x] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [x] I have checked for existing issues that describe my problem prior to opening this one.
- [x] I understand that improperly formatted bug reports may be closed without explanation.

gaby:
## Summary
- Add support for DisableStartupCheck to several storage drivers.

We can add this feature to other drivers in the future if needed.

Fixes #1924

## Summary by CodeRabbit

* **New Features**
  * Added DisableStartupCheck across Memcache, Postgres, Redis and Rueidis to optionally skip startup connectivity checks and resets (default: false).
* **Bug Fixes**
  * Rueidis: avoid panics on uninitialized clients by deferring initialization errors and making Close/operations safe when startup checks are disabled.
* **Documentation**
  * Updated config docs to include DisableStartupCheck and its default.
* **Tests**
  * Added tests verifying stores construct and close without startup checks enabled.


ReneWerner87:
@gaby can you update and check ai hints

gaby:
> @gaby can you update and check ai hints

I don't have the task anymore for this. We can probably re-implement it using copilot.

Closing this one in the meantime.

PR #2695, opened 2026-05-09
🧹 chore: Add unit tests to ensure CockroachDB compatibility

ademdgn:
## Summary by CodeRabbit

* **New Features**
  * CockroachDB is now supported as a database backend, with the existing PostgreSQL driver working seamlessly when connected to a CockroachDB instance.

* **Tests**
  * Added comprehensive integration test suite for CockroachDB, validating core operations including data storage, retrieval, deletion, reset, expiration handling, garbage collection, and connection management.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gofiber/storage/pull/2695)


gaby:
@ademdgn This driver is just a rename. It is using the same library as https://github.com/gofiber/storage/blob/main/postgres/postgres.go#L14

What I think we can do is update the `postgres` driver to also test with coachroachdb and also update the docs.

ademdgn:
@gaby  Thanks for the feedback! You're right, since CockroachDB is wire-compatible with PostgreSQL and uses the same library, a separate driver is redundant.

I've updated the approach:

Added CockroachDB tests to the existing postgres driver
Updated the README to mention CockroachDB compatibility
Removed the separate cockroachdb/ directory
Let me know if anything else needs to be changed!

gaby:
@ademdgn Update .github/workflows/test-postgres.yml with an ENV for the Docker image

gaby (approved):

PR #3346, opened 2026-09-04
chore(deps): updates mongodb driver to v2

justledbetter:
With this update, clients using `go.mongodb.org/mongo-driver/v2` are able to use the Session driver again.  (Renamed package to `github.com/gofiber/storage/mongodb/v3` to preserve compatibility with existing codebases.)

Happy to correct anything I did wrong :)


ReneWerner87:
@justledbetter thanks for picking this up. Here is what is still open, based on the two failed runs and a local build of the branch at `cb8b7536`.

### 1. `mongodb_test.go` was not migrated (blocking)

The test file still imports the v1 `primitive` package, so the package does not build:

```
./mongodb_test.go:362:16: cannot use primitive.NewObjectID() (value of array type primitive.ObjectID) as bson.ObjectID value in assignment
FAIL	github.com/gofiber/storage/mongodb/v3 [build failed]
```

Seen in [Tests Mongodb](https://github.com/gofiber/storage/actions/runs/33911387370) and [Benchmark](https://github.com/gofiber/storage/actions/runs/33911387760), and still reproducible on the current head. Three spots:

- line 14: `go.mongodb.org/mongo-driver/bson/primitive` to `go.mongodb.org/mongo-driver/v2/bson`
- line 362: `primitive.NewObjectID()` to `bson.NewObjectID()`
- line 370: `primitive.ObjectID{}` to `bson.ObjectID{}`

### 2. The v1 driver is still a direct dependency

`go.mongodb.org/mongo-driver v1.17.9` is still in the require block of `mongodb/go.mod`. Once the test file is migrated, `go mod tidy` drops it together with the then unused indirect `github.com/montanaflynn/stats v0.12.5`, and removes the matching 4 lines from `go.sum`. No other go.sum change is needed, since the v2 driver already arrives through `testcontainers-go/modules/mongodb`.

### 3. `mongodb/README.md` was not updated

These files are synced to the documentation site by `sync-docs.yml`, so the README ships as is:

- line 43: `go get github.com/gofiber/storage/mongodb/v2` to `/v3`
- line 49: `import "github.com/gofiber/storage/mongodb/v2"` to `/v3`
- line 142: same import inside the "Using an Existing MongoDB Connection" example, to `/v3`
- lines 143 and 144: `go.mongodb.org/mongo-driver/mongo` and `.../mongo/options` to `go.mongodb.org/mongo-driver/v2/mongo` and `.../v2/mongo/options`
- line 148: `mongo.Connect(context.Background(), options.Client().ApplyURI(...))` to `mongo.Connect(options.Client().ApplyURI(...))`. v2 dropped the context parameter, so the documented example would not compile as written. The `context` import stays, it is still used by `client.Disconnect`.

### 4. Stale doc comment (minor)

`mongodb.go:56` still describes ctx as the parent context for "connect, ping, optional drop, index creation". With v2, `mongo.Connect` no longer takes a context, so connect is no longer bounded by it. Removing "connect" from that list keeps the comment accurate.

### For the record, these parts look right

The rename to `/v3` is correct: `NewFromConnection`, `NewFromConnectionWithContext` and `Conn` expose driver types in the public API, so the driver major bump is a breaking change for consumers. This matches the existing `postgres/v3` and `redis/v3` modules, and the release tag would be `mongodb/v3.0.0`. Dropping the `withDefaultTimeout` wrapper around `mongo.Connect` is also correct, since v2 performs no blocking I/O there and the ping right after is still bounded by ctx.

One note on process: `Tests Mongodb`, `Benchmark` and `Linter` are currently sitting on `action_required` for this head, so they need a maintainer to approve the workflow run before any of this is verified by CI.

justledbetter:
I've made the changes you suggested - Thanks so much for steering me in the right direction!  Hopefully things look better now :smile:

ReneWerner87 (approved):

